PR #16, opened 2023-04-30
Using spot.yml & SPOT_FILE in code

bessarabov:
In docs there is stated:

 * `Defaults to spot.yml`
 * `You can also set the environment variable $SPOT_FILE` 
 
 ( https://github.com/umputun/simplotask/blob/7db4a2cff0c8b00101997f1ae3fcdc9d4539d604/README.md?plain=1#LL43C60-L43C82 ) 
 
 But it is not true in the code. This pull request corrects the code to aligned with the documentation.

### DIFF
--- a/app/main.go
+++ b/app/main.go
@@ -23,7 +23,7 @@ import (
 )
 
 type options struct {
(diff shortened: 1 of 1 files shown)


umputun:
Yep, typo. thx for the fix

PR #18, opened 2023-04-30
Inventory groups

umputun:
This PR adds support of inventory groups. Groups are optional, and "plain" (groupless) files without still supported

i.e. the oll format like this, sill valid:

```text
h1.example.com 
h2.example.com:2222
h3.example.com:2223 user2
```

in addition, ini-style groups can be defined:

```text
[gr1]
h1.example.com 
h2.example.com:2222

[gr2]
h3.example.com:2223 user2
```

PR also changes the inventory format for both `inventory_file` and `inventory_url` to support groups, i.e.

`inventory_file: {location: "testdata/inventory", groups: ["staging"]}`

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,9 +73,11 @@ targets:
   prod:
     hosts: ["h1.example.com", "h2.example.com"]
   staging:
-    inventory_file: "testdata/inventory"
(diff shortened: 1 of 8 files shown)


umputun:
Note: this is a breaking change

PR #25, opened 2023-05-01
Tasks list

umputun:
This PR is breaking. It converts tasks from map to list and introduces a new mandatory field, `name.` Lack of `name` or duplicated names will fail spot execution right away

In addition, it adds `name` to targets. This names have nothing to do with names above, they are just to produce better log output


umputun:
see #23 for the reasoning

Issue #54, opened 2023-05-02
Alternative lightweight  playbook format


umputun:
I have created several playbooks today, and most of them are trivial. They have a single target (usually group and names) and a a single task

The idea is to support some simplified format to allow something not as rich as a complete playbook but more flexible compared to ad-hoc commands.

umputun:
note: this is not a replacement for the current format, but rather additional one to support

umputun:
Probably this one is too simple to be practical. I think we still need a top-level task with multiple commands, i.e.

```yml
task:
   - name: copy something
     copy: ....
   - name: sync something
     sync: ....
  - run something
     script: ....
````

umputun:
The branch `simple-playbook` implements the described above (see #55 for more details). I have covered it with tests, and it should be working. The change is back-compatible.

You can check it out if you like.

umputun:
update: merger to master

PR #55, opened 2023-05-03
Simplfyed  playbook

umputun:
The simplified playbook format offers an alternative to the full playbook and should be easier to read and write. This is **not a replacement** for the regular playbook; internally, it is converted to a full playbook with a single preset target "default" and a single task in the tasks list. 

This is the format:

```yaml
user: umputun                       
ssh_key: keys/id_rsa           
inventory: /etc/spot/inventory.yml  
target: ["devbox1", "devbox2", "h1.example.com:2222", "h2.example.com"] 

task:
  - name: wait
    script: sleep 5s
  
  - name: copy configuration
    copy: {"src": "testdata/conf.yml", "dst": "/tmp/conf.yml", "mkdir": true}
  
  - name: sync things
    sync: {"src": "testdata", "dst": "/tmp/things"}
  
  - name: some command
    script: |
      ls -laR /tmp
      du -hcs /srv
      cat /tmp/conf.yml
      echo all good, 123
  
  - name: delete things
    delete: {"path": "/tmp/things", "recur": true}
  
  - name: show content
    script: ls -laR /tmp

  - name: docker pull and start
    script: |
      docker pull umputun/remark42:latest
      docker stop remark42 || true
      docker rm remark42 || true
      docker run -d --name remark42 -p 8080:8080 umputun/remark42:latest
    env: {FOO: bar, BAR: qux} # set environment variables for the command
    
  - wait: {cmd: "curl -s localhost:8080/health", timeout: "10s", interval: "1s"} # wait for health check to pass
```

Main differences between the two types of playbooks:

- The full playbook supports multiple target sets, while the simplified playbook supports only one target set. I.e. the full playbook can execute the same set of commands on multiple environments where each environment is defined as a separate target set. The simplified playbook can execute the same set of commands only in a single environment.
- The full playbook supports multiple tasks, while the simplified playbook supports only one task. I.e. the full playbook can execute multiple sets of commands, while the simplified playbook can execute only one set of commands.
- The full playbook support various targets type, i.e. `hosts`, `groups`, and `names`, while the simplified playbook supports only one type which is a list of names or host address. 
- Simplified playbook doesn't support task-level `on_error`, `user`, and `ssh_key` fields, while the full playbook does. 

The rest of the fields and options are supported by both types of playbooks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,18 +170,18 @@ task:
 
 ```
 
-## Playbook types
+## Playbook Types
 
-Spot supports two types of playbooks full and simplified. Each one can be represented in even yaml or toml format. The full playbook is more powerful and flexible, but also more verbose and complex. The simplified playbook is easier to read and write, but also more limited in its capabilities.
+Spot supports two types of playbooks: full and simplified. Both can be represented in either YAML or TOML format. The full playbook is more powerful and flexible but also more verbose and complex. The simplified playbook, on the other hand, is easier to read and write but has more limited capabilities.
 
-Main differences between the two types of playbooks:
+Here are the main differences between the two types of playbooks:
 
-- The full playbook supports multiple targets sets, while the simplified playbook supports only one target set. I.e. the full playbook can execute the same set of commands on multiple environments where each environment is defined as a separate target set. The simplified playbook can execute the same set of commands only on a single environment.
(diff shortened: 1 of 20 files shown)


umputun:
this PR address #54

PR #75, opened 2023-05-09
Fix simplified targets key in readme

themagic314:
The simplified readme example has a minor typo which always results to "playbook loaded with 0 tasks"


umputun:
>  Adding a port to targets seems like fixing this. Should we use 22 if no port is specified?
targets: ["10.10.9.21:22"]

this is expected, but not obvios. Probably should be explained better in readme. In case of simplified playbook targets may contain either direct host address:port, or name matching inventory. Without user providing port directly we have no way to understnatd if this is host address or the name.

One improvement i can think of - in case if inventory is not set in the playbook and not passed via cli/env we can treat everyting in targets as addresses, and in this case it will get defalut port if not set

Issue #98, opened 2023-05-16
Add exclude support to other commands


umputun:
I have added `exclude` support for `sync` command; see https://github.com/umputun/spot/pull/97. This is the most obvious use case for exclusion; however, all other commands that support glob can be extended like this. I.e. copy, delete, mcopy

Gauravvpnd:
I would love to work on this issue can you assign me this?

umputun:
sure, all yours

Gauravvpnd:
> sure, all yours

Thank you i will fix this asap.

umputun:
thx; feel free to reach out if any questions. hopefully, you will find the code easy to read; at least this was one of my goals ;)

umputun:
@Gauravvpnd what's up with this one? Still planning to do it or gave up?

bakurin:
I'm taking over if there is no objection.

PR #110, opened 2023-05-29
Exclude support for delete copy

bakurin:
Implementation of the ticket: Add exclude support to other commands #98

### DIFF
--- a/pkg/executor/local.go
+++ b/pkg/executor/local.go
@@ -60,13 +60,29 @@ func (l *Local) Upload(_ context.Context, src, dst string, opts *UpDownOpts) (er
 		return fmt.Errorf("source file %q not found", src)
 	}
 
-	if opts != nil && opts.Mkdir {
+	var mkdir bool
+	var exclude []string
+
+	if opts != nil {
+		mkdir = opts.Mkdir
+		exclude = opts.Exclude
+	}
+
+	if mkdir {
 		if err = os.MkdirAll(filepath.Dir(dst), 0o750); err != nil {
 			return fmt.Errorf("can't create local dir %s: %w", filepath.Dir(dst), err)
 		}
 	}
 
 	for _, match := range matches {
+		relPath, e := filepath.Rel(filepath.Dir(src), match)
+		if e != nil {
+			return fmt.Errorf("failed to build relative path for %s: %w", match, err)
+		}
+		if isExcluded(relPath, exclude) {
+			continue
+		}
+
 		destination := dst
 		if len(matches) > 1 {
 			destination = filepath.Join(dst, filepath.Base(match))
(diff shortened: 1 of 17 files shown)


pkg/executor/remote.go
```diff
@@ -366,62 +423,86 @@ func (ex *Remote) sftpDownload(ctx context.Context, req sftpReq) error {
 	}
 	defer sftpClient.Close()
 
-	remoteFh, err := sftpClient.Open(req.remoteFile)
+	matches, err := sftpClient.Glob(req.remoteFile)
```
umputun, line 426: the whole change here is misplaced. sftDownload is the wrong abstraction level to perform all of this higher-level matching logic. As you can see, other methods do it on a higher level, i.e. Sync. Delete and so on.

umputun (changes requested):
Thanks. Generally, the PR looks good. The only major issue here is with the sftpDownload changes. First of all, this seems to be the wrong place to do it, and as a result, it has made the method too large and responsible for far too many things.

Another issue is with the tests. This is not a critical one, and you handled it fine, but generally, I would suggest creating new tests (in your case, adding new test data) in a way that doesn't break the existing tests and force you to adjust them, unless it is absolutely necessary.

pkg/executor/remote.go
bakurin: You are right, my bad. I've moved the glob-exclude logic to the level of the Download function.

pkg/executor/local.go
```diff
@@ -255,3 +277,64 @@ func (l *Local) copyFile(src, dst string) error {
...
+
+		return filepath.SkipDir
+	})
+
+	if err != nil {
+		return err
+	}
+
+	if !hasExclusion {
+		return os.RemoveAll(src)
```
umputun, line 336: this part is confusing me a little bit. the walk above removing things too, but probably something left? If this is the case a comment explaining this part will be nice

pkg/executor/remote.go
```diff
@@ -105,15 +118,40 @@ func (ex *Remote) Download(ctx context.Context, remote, local string, opts *UpDo
...
+	}
+
+	remoteFiles, err := ex.findMatchedFiles(remote, exclude)
+	if err != nil {
+		return fmt.Errorf("failed to list remote files by glob for %s: %w", remote, err)
+	}
+
+	for _, remoteFile := range remoteFiles {
+		localFile := local
+		if filepath.Base(remote) != filepath.Base(remoteFile) {
```
umputun, line 136: I don't understand this if, probably a comment will help

umputun (commented):
a couple of questions/suggesting for clarity

pkg/executor/local.go
bakurin: The idea is to remove the top-level directory if there are no actual excursions. For instance, when an `exclude` list is provided but there are no files that matched the items in the list. In such a scenario `walk` is going to bypass the top-level dir `./`.
Does it make sense for you?

pkg/executor/remote.go
bakurin: Here I compare a glob `match` with the `remote` which can be a filename of a glob. Basically, it makes the same as following code snipped does in `Upload` in `Remote` executor. Still, I cannot rely on `len(remoteFiles)` because `exclude` is already applied to the glob result.

```go
remoteFile := remote
if len(matches) > 1 { // if there are multiple files, treat remote as a directory
	remoteFile = filepath.Join(remote, filepath.Base(match))
}
```
But, definitely, it worth it to add a comment.

umputun:
thx, looking good. Can you pls rebase/squash your commits?

Issue #122, opened 2023-06-12
Wrong leading spaces/identation in yaml/script parsing?


knutov:
```
spot --version
spot v1.7.0-33fac8b-2023-06-06T00:48:21Z
```

```yaml
  - name: prom_pass
    commands:
      - name: pass
        script: |
          cat << EOF > /etc/prometheus/prometheus.web.yml
          basic_auth_users:
            admin: $PASS
          EOF
```

```bash
$ cat prom.yml | yq '.tasks[2].commands[0].script' | sed 's/"//g' | sed 's/\\n/\n/g'
```

```
cat << EOF > /etc/prometheus/prometheus.web.yml
basic_auth_users:
  admin: $PASS
EOF
```

```bash
spot -p prom.yml --dbg --dry
```

```
[s2 s2.hidden.tld:22]  ! command script /tmp/.spot/spot-script3829472563
[s2 s2.hidden.tld:22]  > #!/bin/sh
[s2 s2.hidden.tld:22]  > set -e
[s2 s2.hidden.tld:22]  > cat << EOF > /etc/prometheus/prometheus.web.yml
[s2 s2.hidden.tld:22]  > basic_auth_users:
[s2 s2.hidden.tld:22]  > admin: $PASS
[s2 s2.hidden.tld:22]  > EOF
```

There is no leading spaces in `admin: $PASS` line and file created by this script does not contain them too which leads to wrong yaml config.

umputun:
Yeah, this is expected. We are stripping leading spaces intentionally. I have not seen a use case where spaces are essential, but yours indicates the opposite. I'm not sure what the best approach is here - from one hand, we want to strip all the yml-related spaces in most of the cases, but sometimes we need to keep them

knutov:
Can not imagine any reason to strip leading spaces from multiline yaml string. Are there any examples where it's better to do it?

umputun:
oops, edited your comment by accident, planned to reply

umputun:
> I do use a lot of code with `cat << EOF > ` construction with preformatted files. There are number of ways to strip leading spaces in HEREDOC blocks if needed (like `cat <<-EOF`) so why to strip more of code, supposed to be run directly with sh/bash ?

see above. I'm not sure if my cleanup makes much sense in the context you provided, let me think about it and see what would be the negative side effects if we drop the stripping

knutov:
```
 script: |
   cat blah
   echo 123
    ls -la /srv
```

[...]

PR #124, opened 2023-06-12
reference to YAML multiline string formatting examples/docs

knutov:
Add reference to YAML multiline string formatting examples/docs for #122


umputun (approved):
thx

PR #127, opened 2023-07-09
Support env file

umputun:
relatated to https://github.com/umputun/spot/discussions/126

This PR adds basic support for env file in yaml format. The default is `env.yaml` but can be changed in the command line. In case this file is missing, it won't fail and will continue as usual, so the change is back-compatible.

example of env.yml:

```yml
vars:
  var1: val1
  var2: val2
```


umputun:
note: the top level `vars` added intentionally as some day we may want to extend the format to more sophisticated use case

Issue #109, opened 2023-05-28
Plugin for IDEA/VScode


umputun:
It would be very nice to have a plugin to help work with spot's playbook files and inventory. In addition to verifying the validity, it should also be capable of running tasks

hnkisdead:
I tried to write json schema for inventory file and in IDEA-like IDEs it works fine for autocomplete but validation is strange. 

Take for example validation for host field. For whatever reason this validation works if both `groups` and `hosts` is used. But schema define them as mutually exclusive and IDE allow using both of them.

<img width="259" alt="Screenshot 2023-07-08 at 13 52 47" src="https://github.com/umputun/spot/assets/5451490/7d0e8baf-8e60-4e16-98f3-aadca7d1e7b6">
<img width="390" alt="Screenshot 2023-07-08 at 13 52 57" src="https://github.com/umputun/spot/assets/5451490/748bb938-e3d1-4f85-bbd7-3b36e1919823">

I will check schema with other tools for validation and maybe they give some insights where is problem: in schema definition or in IDE

Schema:
[spot-inventory.json.gz](https://github.com/umputun/spot/files/11990279/spot-inventory.json.gz)

PS: i also tried write plugin for IDEA, its hard but doable, but i can't predict how long it takes, cause i never done it before.

umputun:
I have tried to generate schema with chatgpt, and the result passes [this validator](https://www.jsonschemavalidator.net), however, the IDEA still complains about it
![image](https://github.com/umputun/spot/assets/535880/b3468c80-3446-47e5-9a87-45dad7854271)

Not sure what is wrong with this:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "$id": "spot-inventory.json",
  "title": "spot inventory file",
  "type": "object",
  "anyOf": [
    {
      "required": ["hosts"]
    },
    {
      "required": ["groups"]
    }
  ],
  "properties": {
    "hosts": {
      "type": "array",
      "items": {
        "$ref": "#/definitions/host"
      }
    },
    "groups": {
      "type": "object",
      "patternProperties": {
        ".*": {
          "type": "array",
          "items": {
            "$ref": "#/definitions/host"
          }
        }
      },
      "additionalProperties": false
    }
  },
  "definitions": {
    "host": {
      "type": "object",
      "properties": {
        "name": {
          "type": "string"
        },
        "host": {
          "type": "string"
        },
        "port": {
          "type": "integer",
          "minimum": 1,
[...]

PR #134, opened 2023-07-17
add json schema

hnkisdead:
This PR adds:
- JSON Schemas for inventory and playbook
- README about how to use them in VSCode and IDEA-based IDEs

Issue: https://github.com/umputun/spot/issues/109


umputun:
thx. it would be nice to add a reference to this schema/README.md from the primary README.md. Probably a new section will do

Issue #131, opened 2023-07-17
Indicating multiple task through the command-line options


ifraixedes:
According to the docs `spot` executes all the tasks in the playbook if the `--task` option is passed to the command.

Is there any reason why this option cannot be specified multiple times as `--target, -t` for indicating to execute more than one task?

umputun:
Good point. Technically it is already possible to run multiple tasks (if no `--task` is set); however, it is running all the tasks and not allowing to set the list of tasks.

I don't recall the reason for this; probably, I have not considered this use case as something valid. Either way I think this is a nice thing to have and should be relatively easy to implement.

ifraixedes:
Ok, good.
I may try to take a look at it this week, but I don't promise.

ifraixedes:
Fine to me, I didn't find the time last week, and I doubt that I was going to find it this week.
Anyway, the improvement was nice to have for me because calling `spot` more than one passing the task isn't a problem for me because I need to perform more operations before calling it and I have all of them under a Makefile target.

PR #143, opened 2023-07-25
Multiple task cli option

bakurin:
Support for multiple `--task` CLI option requested in https://github.com/umputun/spot/issues/131

### DIFF
--- a/cmd/spot/main.go
+++ b/cmd/spot/main.go
@@ -32,7 +32,7 @@ type options struct {
 	} `positional-args:"yes" positional-optional:"yes"`
 
 	PlaybookFile string        `short:"p" long:"playbook" env:"SPOT_PLAYBOOK" description:"playbook file" default:"spot.yml"`
-	TaskName     string        `long:"task" description:"task name"`
+	TaskNames    []string      `long:"task" description:"task name"`
 	Targets      []string      `short:"t" long:"target" description:"target name" default:"default"`
 	Concurrent   int           `short:"c" long:"concurrent" description:"concurrent tasks" default:"1"`
 	SSHTimeout   time.Duration `long:"timeout" env:"SPOT_TIMEOUT" description:"ssh timeout" default:"30s"`
@@ -154,7 +154,7 @@ func run(opts options) error {
 		return runGen(opts, r)
 	}
 
-	if err := runTasks(ctx, opts.TaskName, opts.Targets, r); err != nil {
+	if err := runTasks(ctx, opts.TaskNames, opts.Targets, r); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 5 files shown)


cmd/spot/main.go
```diff
@@ -163,12 +163,14 @@ func run(opts options) error {
...
 // runTasks runs all tasks in playbook by default or a single task if specified in command line
-func runTasks(ctx context.Context, taskName string, targets []string, r *runner.Process) error {
-	// run a single task if specified
-	if taskName != "" {
-		for _, targetName := range targetsForTask(targets, taskName, r.Playbook) {
-			if err := runTaskForTarget(ctx, r, taskName, targetName); err != nil {
-				return err
+func runTasks(ctx context.Context, taskNames, targets []string, r *runner.Process) error {
+	// run specified tasks if there is any
+	if len(taskNames) > 0 {
```
umputun, line 168: i think this condition can be removed as the for below won't do anything on the empy list anyway

cmd/spot/main.go
```diff
@@ -198,7 +200,10 @@ func runAdHoc(ctx context.Context, targets []string, r *runner.Process) error {
 
 // runGen generates a destination report for the task's targets
 func runGen(opts options, r *runner.Process) (err error) {
-	targets := targetsForTask(opts.Targets, opts.TaskName, r.Playbook)
+	var targets []string
+	for _, taskName := range opts.TaskNames {
```
umputun, line 204: probably this will cause duplicated  targets in the final list. I'm not sure how bad the end result is going to be but deduplication of the result slice sounds like the right thing to do

umputun (changes requested):
thx. just a coule of minor things

cmd/spot/main.go
bakurin: There is a return inside this if statement which stops function execution if tasks are specified and before running all tasks in the playbook. So I think this condition is needed here. Or I misunderstood something?

cmd/spot/main.go
bakurin: Actually, it causes duplicates in the targets list. I've updated the code and added a test to cover the case.

cmd/spot/main.go
```diff
@@ -198,7 +200,17 @@ func runAdHoc(ctx context.Context, targets []string, r *runner.Process) error {
 
 // runGen generates a destination report for the task's targets
```
ifraixedes, line 201: nit: "task's targets" to "tasks' target"

bakurin: Thanks for your input. Fixed.

umputun:
LGTM, can you pls rebase/squash into a single commit, and I'll merge

bakurin:
All changes are squashed and rebased on the latest master.
P.S. Thanks for your review.

umputun (approved):

Issue #170, opened 2024-02-13
Ssh-agent is not called


olegmlsn:
Running a simple playbook 

```yaml
targets: [192.168.234.121:22]

task:
  - name: test
    script: ls -laR /home/empl
```

I do not specify ssh key in playbook or in command line arguments. I specify the `--ssh-agent` option.

```sh
spot -p ../spot_test_data/pb.yml --user=empl --ssh-agent --dbg
```

I expect to see a call to ssh-agent, but I get an error

```sh
spot v1.13.0-72120d5-2024-02-01T20:45:30Z
2024/02/13 15:23:51.101 [DEBUG] request to load playbook "../spot_test_data/pb.yml"
2024/02/13 15:23:51.102 [DEBUG] set target host 192.168.234.121:22
2024/02/13 15:23:51.102 [INFO]  playbook loaded with 1 tasks
2024/02/13 15:23:51.102 [DEBUG] load command "test" (task: default)
2024/02/13 15:23:51.102 [INFO]  no inventory loaded
2024/02/13 15:23:51.102 [INFO]  ssh key: /Users/olegmlsn/.ssh/id_rsa
2024/02/13 15:23:51.102 [DEBUG] use private key "/Users/olegmlsn/.ssh/id_rsa"
2024/02/13 15:23:51.103 [ERROR] can't make runner: can't create connector: private key file "/Users/olegmlsn/.ssh/id_rsa" does not exist
>>> stack trace:
main.main()
        /home/runner/work/spot/spot/cmd/spot/main.go:115 +0x210
panic: [ERROR] can't make runner: can't create connector: private key file "/Users/olegmlsn/.ssh/id_rsa" does not exist

goroutine 1 [running]:
log.Panicf({0x100c3f4a0?, 0x0?}, {0x14000171f18?, 0x18?, 0x0?})
        /opt/hostedtoolcache/go/1.21.6/x64/src/log/log.go:439 +0x64
main.main()
        /home/runner/work/spot/spot/cmd/spot/main.go:115 +0x210
```

PR #171, opened 2024-02-13
Do not select the default key with --ssh=agent

olegmlsn:
should address #170 

If no ssh key is specified and the `--ssh-agent` option is specified, do not select the default key

### DIFF
--- a/cmd/spot/main_test.go
+++ b/cmd/spot/main_test.go
@@ -421,6 +421,21 @@ func Test_sshUserAndKey(t *testing.T) {
 			expectedKey:  filepath.Join(osUser.HomeDir, ".ssh", "id_rsa"),
 		},
 		{
+			name: "SSHAgent set no key in playbook and command line",
+			opts: options{
+				TaskNames: []string{"test_task"},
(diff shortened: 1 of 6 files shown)


cmd/spot/main_test.go
```diff
@@ -441,7 +441,7 @@
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			key, err := sshKey(tc.opts.SSHKey, &tc.conf)
+			key, err := sshKey(tc.opts.SSHAgent, tc.opts.SSHKey, &tc.conf)
```
umputun, line 444: would be nice to add another test case above with opts.ssh-agent set

umputun (commented):
thx. a small thing what is missing - a small test case for the change

olegmlsn:
Evil linter ). I couldn't beat him. Nolint?
```go
//nolint:revive //Evil linter
```

umputun:
re: linter thing - I have updated master, pls rebase your branch and it should pass

umputun:
LGTM, thx

Issue #178, opened 2024-03-21
Add support of env and secrets to onerror action


umputun:
### Discussed in https://github.com/umputun/spot/discussions/177

<div type='discussions-op-text'>

<sup>Originally posted by **rahfar** March 21, 2024</sup>
Hi, I want to add notification to telegram on error. Do not want to hardcode bot token in playbook, is it possible to use env vars or secrets in on_error field?</div>

rahfar:
I assume it could be closed - at least, I am fully satisfied with the new availability of setting secrets, and environment variables at the task level. thanks

PR #179, opened 2024-03-21
extend support of env vars and secrets for onerror command #178

umputun:
see discussion #177 

The change is propagating environment setting as well as secrets from the task triggered onerror call. This is follow the standard way we treat those in `script` command and should be working with both a single line and multiline forms of `onerror`

### DIFF
--- a/pkg/runner/commands.go
+++ b/pkg/runner/commands.go
@@ -40,8 +40,8 @@ type execCmdResp struct {
 }
 
 type execCmdErr struct {
-	err error
-	cmd execCmd
+	err  error
+	exec execCmd
 }
 
 func (e execCmdErr) Error() string { return e.err.Error() }
@@ -65,7 +65,7 @@ func (ec *execCmd) Script(ctx context.Context) (resp execCmdResp, err error) {
 	single, multiRdr := ec.cmd.GetScript()
 	c, scr, teardown, err := ec.prepScript(ctx, single, multiRdr)
 	if err != nil {
-		return resp, &execCmdErr{err: fmt.Errorf("can't prepare script on %s: %w", ec.hostAddr, err), cmd: *ec}
+		return resp, &execCmdErr{err: fmt.Errorf("can't prepare script on %s: %w", ec.hostAddr, err), exec: *ec}
 	}
 	defer func() {
 		if teardown == nil {
@@ -96,7 +96,7 @@ func (ec *execCmd) Script(ctx context.Context) (resp execCmdResp, err error) {
 
 	// collect setvar output to vars and latter it will be set to the environment. This is needed for the next commands.
 	// setenv output is in the format of "setenv foo=bar" and it is appended to the output by the script itself.
-	// this part done inside cmd.scriptFile function.
+	// this part done inside exec.scriptFile function.
 	resp.vars = make(map[string]string)
 	for _, line := range out {
 		if !strings.HasPrefix(line, "setvar ") {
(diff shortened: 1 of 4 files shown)


umputun:
@rahfar this PR adds support for both. If you have a chance, please give it a try from the `onerror_extras`  branch. If you need help with creating the binary, let me know for which platform, and I'll build it for you.

I have covered this change with a separate test using https://github.com/umputun/spot/blob/onerror_extras/pkg/runner/testdata/conf-onerror.yml, which you can use as a reference.

rahfar:
great, many thanks!

I have tested with setting webhook URL in `env.yml` and it works just fine.

With secrets it is more unclear to me from where does it take? In the provided example error occures in the command that has secret option specified and those secrets are available in on_error step. However, what happens if an error occurs in a command that does not have the secrets option? Will the secrets still be visible to the on_error command? If not, that means that I should specify secrets option in every command, which would significantly worsen the readability of the playbook

rahfar:
Sorry for offtop - I do not know if such behaviour should be another issue:
if local command fails - in logs there is message about error connecting to default:22 instead of desc of failed command
despite the error code is correct
```
spot latest
[localhost] run task "fail-task", commands: 1 (local)
failed, can't run task "fail-task" for target "default": 1 error(s) occurred:
   [0] failed command "fail-cmd" on host default:22 (): can't run script on default:22: exit status 2
```

playbook
```
user: root
ssh_key: ~/.ssh/id_rsa
ssh_shell: /bin/bash
local_shell: /bin/bash

tasks:
  - name: fail-task
    commands:
      - name: fail-cmd
        script: "ls nosuchdir"
        options:
          local: true
```

umputun:
> With secrets it is more unclear to me from where does it take? 

From the secret's provider, see [docs](https://github.com/umputun/spot?tab=readme-ov-file#secrets) for more details.

> In the provided example error occurs

This is not truly an example, but rather a test. This test is running with a secret provider, and as long as you pick yours and put a secret in, it will work the same way.

>  what happens if an error occurs in a command that does not have the secrets option? 

Secrets are only injected into the system if the command directly requests it. This is intentional behavior that is unlikely to be changed. Theoretically, adding secrets directly in `onerror` will meet your use case better, but  I don't see a way of doing it yet without breaking compatibility. 

Probably, a simple trick of defining secrets in the first command and setting them to the environment will do what you want and this way all your commands will get the secret

umputun:
> if local command fails - in logs there is message about error connecting to default:22 

It doesn't say anything about "connecting". However, this default:22 part seems confusing. This happens because the message is built from a common pattern for both remote and local commands. I'll see if it can be quickly fixed

rahfar:
> Secrets are only injected into the system if the command directly requests it. This is intentional behavior that is unlikely to be changed. Theoretically, adding secrets directly in onerror will meet your use case better, but I don't see a way of doing it yet without breaking compatibility.

Got it.
I am using spot secret provider and also thought about adding same configuration schema to on_error as for command.
But yeah totally agree that it would ruin compatibility 

> Probably, a simple trick of defining secrets in the first command and setting them to the environment will do what you want and 
this way all your commands will get the secret

thanks for the workaround!

rahfar:
> It doesn't say anything about "connecting". However, this default:22 part seems confusing. This happens because the message is built from a common pattern for both remote and local commands. I'll see if it can be quickly fixed

thanks! 
As a workaround explicitly setting `targets: ["localhost"]` makes log looking better:
```
spot latest
[localhost] run task "fail-task", commands: 1 (local)
failed, can't run task "fail-task" for target "localhost": 1 error(s) occurred:
   [0] failed command "fail-cmd" on host localhost:22 (): can't run script on localhost:22: exit status 2
```

umputun:
Regarding applying secrets to the whole task without setting it per command - see #180 

This solution seems sound and solves the issue you mentioned as a nice side-effect of populating default options (including secret keys) from the task to all the commands.

rahfar:
Awesome!

PR #186, opened 2024-03-24
Add system env variable expansion to envVars function

umputun:
The function `envVars` has been updated to enable the application to also read system environment variables, in addition to CLI and env file variables. This change includes an update of the function's comment to reflect the new behavior, and the addition of new test cases to validate the feature. CLI variables override env file variables, which in turn override system environment variables if they share the same name.

see discussion #184


umputun:
@rahfar - this PR adds support for the OS environment. Give it a try (merged to the master), and let me know if it works for you.

to clarify: in both CLI and env file you can use smth like `-e var1=$HOME` (also `${HOME}`) or `- var1: $HOME`

rahfar:
tried it, and there are 2 points that holds me to use it instead of `load_secrets_to_spot.sh && spot; rm spot.db`:
1. env vars are printed in logs
2. (not very important) would like to have `MYSECRET: $MYSECRET` inside playbook, instead of in cli arg or env.yml

but in general, it is a good feature

PR #205, opened 2024-05-18
Use local executor if hostAddr is local

koolay:


umputun (commented):
thx for the PR. However, there are a few issues here:

- Technically, the check you do for the localhost is way too broad. It will catch anything having "localhost" in it, like "localhost.mydomain.com".
I don't think you should even handle 127.0.0.1 because it is at best a partial solution. Keeping just localhost as a special case would be better.
- The change is lacking any corresponding tests.
- The change is missing a readme/docs update.

In the [discussion we had](https://github.com/umputun/spot/discussions/204), I have emphasized why I do not like this approach, and you have not yet convinced me that this is even a good idea. However, if the final implementation will be as simple as it looks like now, I may approve it if the points above are addressed.

PR #218, opened 2024-06-30
Use slog

koolay:
Use slog instead of log


umputun:
I don't see a good reasons doing it. Theoretically a reason can be to get rid of dependency if the stdlib-based implementation provides comparable functionality, but it doesn't. For example all the colorization part is completely missing.

Another issue here - the change seems to be a combination of a technical find/replace with an additional reformatting of completely unrelated code. Both parts are not problematic - `slog.Debug(fmt.Sprintf("run %s", cmd))` doesn't need Sprintf, and code reformatting has nothing to do with this change and not needed.

PR #236, opened 2024-12-12
Adds support for SSH ProxyCommand to be able to use "Bastion" servers

vo-va:
Adds support for this https://man.openbsd.org/ssh_config.5#ProxyCommand

Discussion https://github.com/umputun/spot/discussions/231 

Documentation not updated yet. I will update it if implementation is ok and will not require changes. Please check comments in connector_test.go. 

I was running tests on a local computer, some of which related to vaults were failing for me. If they fail on GitHub, I will figure this out later.


umputun:
looks like some minor linter issues

vo-va:
I've tried to use linter before pushing a new commit, but I think I do not fully understand why it was complaining about Shadows declaration inside the if block and not complaining about err that is declared for agent forwarding. Idea is reporting it, so I am not sure it was fully fixed. Looks like name return of sshClient() is affecting it.

umputun:
I can't even allow this pipeline to run anymore. GH supposed to request permission to activate the pipeline, but it shows nothing. Anyway, don't worry about it. I'll review the code today (hopefully) regardless of those linter quirks.

pkg/config/playbook.go
```diff
@@ -78,11 +78,12 @@ type Target struct {
...
-	Host string   `yaml:"host" toml:"host"`
-	Port int      `yaml:"port" toml:"port"`
-	User string   `yaml:"user" toml:"user"`
-	Tags []string `yaml:"tags" toml:"tags"`
+	Name         string   `yaml:"name" toml:"name"`
+	Host         string   `yaml:"host" toml:"host"`
+	Port         int      `yaml:"port" toml:"port"`
+	User         string   `yaml:"user" toml:"user"`
+	Tags         []string `yaml:"tags" toml:"tags"`
+	ProxyCommand []string `yaml:"proxy_command" toml:"proxy_command"`
```
umputun, line 86: I'm not sure if having []string here is clear and user-friendly. I would expect a single string with all the arguments.

pkg/executor/connector.go
```diff
@@ -22,6 +23,31 @@ type Connector struct {
 	logs                  Logs
 }
 
+// SubstituteProxyCommand updates variables with values associated with the target host.
+// SSH ProxyCommand can use placeholders such as %h, %p, and %r (%r - username), they have to be replaced with the actual values.
+func SubstituteProxyCommand(username, address string, proxyCommand []string) ([]string, error) {
```
umputun, line 28: Is this one supposed to be public? Unless I missed something, I don't see why the end user would hit it directly.

pkg/executor/connector.go
```diff
@@ -52,9 +78,9 @@ func (c *Connector) WithAgentForwarding() *Connector {
 }
 
 // Connect connects to a remote hostAddr and returns a remote executer, caller must close.
-func (c *Connector) Connect(ctx context.Context, hostAddr, hostName, user string) (*Remote, error) {
-	log.Printf("[DEBUG] connect to %q (%s), user %q", hostAddr, hostName, user)
-	client, err := c.sshClient(ctx, hostAddr, user)
+func (c *Connector) Connect(ctx context.Context, hostAddr, hostName, user string, proxyCommand []string) (*Remote, error) {
```
umputun, line 81: I don't like changing the connect method to add an optional argument. I would rather expect another WithProxy method, similar to WithAgent.

pkg/executor/connector.go
```diff
@@ -52,9 +78,9 @@ func (c *Connector) WithAgentForwarding() *Connector {
...
+	log.Printf("[DEBUG] connect to %q (%s), user %q, proxy command: %s", hostAddr, hostName, user, proxyCommand)
```
umputun, line 82: As proxy is very optional, I won't add it to the default debug message(s). Probably anything regarding the proxy should be shown in debug only if this proxy is active.

pkg/executor/connector.go
```diff
@@ -89,27 +115,73 @@ func (c *Connector) forwardAgent(client *ssh.Client) error {
...
 
 	conf, err := c.sshConfig(user, c.privateKey)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create ssh config: %w", err)
 	}
-	ncc, chans, reqs, err := ssh.NewClientConn(conn, host, conf)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create client connection to %s: %v", host, err)
+
+	if len(proxyCommand) == 0 {
```
umputun, line 160: Looks complicated to me; it probably deserves to be extracted to a separate function.

umputun (changes requested):
Thanks for the PR. I have indicated a few issues I have with this PR. Another thing I would like to see is more tests to ensure things work as expected, end-to-end. At least a main-level test (main_test.go), but generally as much testing as needed to be absolutely certain it works and doesn't break anything existing.

vo-va:
Just in case, I am working on addressing review comments. It is not abandoned.

pkg/executor/connector.go
vo-va: May I ask about this place https://github.com/umputun/spot/blob/master/pkg/runner/runner.go#L187  I guess there will be a 2nd function Connect.ConnectWithProxy(...), but what approach should I use to choose between normal connection and proxy? Will `if ... else` be good enough here?

pkg/executor/connector.go
vo-va: To be sure that I am not making changes in the wrong direction, I want to ask next. If the optional parameter ProxyCommand will be configured with a method similar to .WithAgent(), the code that establishes a connection through ProxyCommand will be a separate method, but at some place I will use `if … else` or just `if … return, if … return` to choose between a normal network connection and a proxy connection, will it be acceptable?

vo-va:
@umputun Hi, may I ask you to check my questions under your review comments

vo-va:
Hi, please check the updated pull request.

I think there are too many changes, so you’ll probably request more. I’ve added some comments in the code to explain why they were done this way. The main challenge I found is that the --target CLI argument can override the host in a quite deep place in the code, so to support it I had to make many changes.

I was thinking about moving that code up somewhere else, but I have a hunch it wouldn’t be accepted.

image


The linter reports some issues. The warning about possible tainted input seems to be a false positive because the code needs to run an external command.

The report about testcontainers.NetworkRequest being deprecated — I didn’t try too hard to fix it, since it looks like it would require updating packages and modifying the vendor directory. I expect those changes wouldn’t be accepted.

Test SSH key files changed permissions because in tests the code uses the system SSH client, to establish tunnel, which requires the key to have permissions only for the current user.

PR #312, opened 2025-07-30
Issue 311 support task tagging

supermunn:
This change adds ability to tag tasks. 

In the playbook, you can add `tags` to each task:

```
---
user: ubuntu
ssh_key: "<PATH_TO_SSH_KEY>"
ssh_shell: "/bin/bash"
local_shell: "/bin/bash"
inventory: "spot_inventory.yaml"

tasks:
- name: My tagged task
  tags: ['my-tag']
  commands:
    - name: Run the check and store result
      script: 'echo "hello moto"'
      options:
        sudo: true
        ignore_errors: true
  targets:
    - database_primary_node
```

To run tasks with matching tags: 

```
spot -p <PATH_TO_PLAYBOOK_YAML> --tags=my-tag 
```

To skip tasks with matching skip-tags:

```
spot -p <PATH_TO_PLAYBOOK_YAML> --skip-tags=my-test 
```

### DIFF
--- a/cmd/spot/main.go
+++ b/cmd/spot/main.go
@@ -42,6 +42,10 @@ type options struct {
 	SSHShell        string        `long:"shell" env:"SPOT_SHELL" description:"enforce non-default shell to use for ssh" default:""`
 	SSHTempDir      string        `long:"temp" env:"SPOT_TEMP" description:"temporary directory for ssh" default:""`
 
+	// tasks filter
+	Tags     []string `long:"tags" description:"only run tasks that match one or more of the specified tags"`
+	SkipTags []string `long:"skip-tags" description:"skip tasks that match one or more of the specified tags"`
+
 	// overrides
 	Inventory string            `short:"i" long:"inventory" description:"inventory file or url [$SPOT_INVENTORY]"`
 	SSHUser   string            `short:"u" long:"user" description:"ssh user"`
@@ -355,6 +359,8 @@ func makeRunner(opts options, pbook *config.PlayBook) (*runner.Process, error) {
 		Dry:         opts.Dry,
 		SSHShell:    opts.SSHShell,
 		SSHTempDir:  opts.SSHTempDir,
+		Tags:        opts.Tags,
+		SkipTags:    opts.SkipTags,
 	}
 	log.Printf("[DEBUG] runner created: concurrency:%d, connector: %s, ssh_shell:%q, verbose:%v, dry:%v, only:%v, skip:%v",
 		r.Concurrency, r.Connector, r.SSHShell, r.Verbose, r.Dry, r.Only, r.Skip)
--- a/pkg/config/playbook.go
+++ b/pkg/config/playbook.go
@@ -67,7 +67,7 @@ type Task struct {
 	OnError  string     `yaml:"on_error" toml:"on_error"`
 	Targets  []string   `yaml:"targets" toml:"targets"`           // optional list of targets to run task on, names or groups
 	Options  CmdOptions `yaml:"options" toml:"options,omitempty"` // options for all commands
-        Tags     []string    `yaml:"tags" toml:"tags"` // optional list of tags to add to tasks
+	Tags     []string   `yaml:"tags" toml:"tags"`                 // optional list of tags to add to tasks
 }
 
 // Target defines hosts to run commands on
--- a/pkg/config/playbook.go
+++ b/pkg/config/playbook.go
@@ -67,6 +67,7 @@ type Task struct {
 	OnError  string     `yaml:"on_error" toml:"on_error"`
 	Targets  []string   `yaml:"targets" toml:"targets"`           // optional list of targets to run task on, names or groups
 	Options  CmdOptions `yaml:"options" toml:"options,omitempty"` // options for all commands
+        Tags     []string    `yaml:"tags" toml:"tags"` // optional list of tags to add to tasks
 }
 
 // Target defines hosts to run commands on
--- a/pkg/runner/runner.go
+++ b/pkg/runner/runner.go
@@ -84,7 +84,7 @@ func (p *Process) Run(ctx context.Context, task, target string) (s ProcResp, err
 		return ProcResp{}, fmt.Errorf("can't get task %s: %w", task, err)
 	}
 
-	if !p.shouldRunTask(tsk) {
+	if !p.shouldRunTask(*tsk) {
 		log.Printf("[runner] skipping task %q due to tag filtering", tsk.Name)
 		return ProcResp{}, nil
 	}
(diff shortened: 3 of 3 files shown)


umputun:
Hi @supermunn, thanks for submitting the PR\! I've reviewed your implementation and have some thoughts to share.

**The Good:**
- Your code implementation looks reasonable and follows the approach we discussed in the issue
- The task filtering logic is clean and well-structured
- I see this is marked as draft and missing tests/docs, which is perfectly fine for a work-in-progress

**Conceptual Consideration:**
I've been thinking about the introduction of `--tags` and `--skip-tags` flags, and I have a concern about the conceptual model. Both `--task` and `--tags` are essentially task selection mechanisms, which creates a potential confusion:
- `--task=deploy` - run specific task by name
- `--tags=prod` - run tasks with specific tags

Having both together (e.g., `spot -n deploy --tags=prod`) creates ambiguity about what we're trying to achieve.

**Alternative Approach:**
Looking at how Spot handles target selection, we use a smart resolution approach. The `-t` flag can accept:
- Target names from playbook
- Group names from inventory  
- Tags from inventory
- Host names or addresses

The system figures out what you mean based on priority order (target def → group → tag → host → address).

We could apply the same pattern to tasks. Instead of adding new flags, the existing `-n/--task` flag could accept both task names and tags:
```bash
spot -n deploy        # Runs task "deploy" if it exists, OR all tasks tagged "deploy"
spot -n prod -n test  # Mix of task names and tags, resolved intelligently
```

This would:
- Eliminate the conceptual conflict
- Keep the CLI simpler (no new flags)
- Be consistent with existing Spot patterns
- Maintain backwards compatibility

What do you think about this approach? Happy to discuss further if you'd like to explore this direction.

supermunn:
hi @umputun thank you so much for your feedback. I wasn't aware of the intelligent target resolution in Spot, glad to know that mechanism exists. 

My personal view on the alternative approach is that, I personally prefer CLI flags that are clear and separated; ie `-n` is only for selecting task name, `--tags` is for tags. But that's my personal preference. I do understand the wish to keep it consistent with how Spot currently is. 

**Question:** I wanted to ask you another option which is commonly used in cli commands. How do you feel about ignoring `--tags` if `-n` is set?

supermunn:
hey @umputun 👋 I made some changes given your feedback hoping that I put my changes in the correct place. Ran the unit test and also tested with a playbook. Before I mark it ready, would you be able to have a look? Thank you so much, this provided me with a great opportunity to get familiar with the Spot code.

umputun:
Since you mentioned this is a great learning opportunity and you're getting familiar with the Spot codebase, I've prepared a more detailed review than usual. This might be overly detailed for a typical PR review, but I think it will be helpful for understanding both the specific issues and the general Go/Spot patterns. Feel free to ask questions if anything isn't clear!

Genereally speaking, the core functionality looks close enought and follows the design we discussed. Here are some areas that need attention:

## Critical Issues

### 1. **Non-deterministic Task Execution Order**
**Problem:** The map-to-slice conversion in `getTaskNamesToRun` produces random task order on each run:
```go
for name := range taskSet {
    result = append(result, name)
}
```

**Why this matters:** Task execution order should be predictable and consistent. Users expect tasks to run in the same order they're defined in the playbook, not randomly.

**Suggested fix:** Preserve the original task order from the playbook when building the result list.

### 2. **Business Logic in Wrong Layer**
**Problem:** The `getTaskNamesToRun` function with complex resolution logic is implemented in `main.go`.

**Why this matters:**
- Violates separation of concerns - `main.go` should orchestrate, not implement business logic
- Makes the code harder to test (need to test through main instead of unit testing the logic directly)
- The PlayBook struct already owns the tasks data, so the logic should live there

**Suggested fix:** Move this logic to a method on the PlayBook struct (e.g., `ResolveTasks`). The PlayBook already has methods like `Task()` and `AllTasks()`, so task resolution naturally belongs there.

### 3. **Silent Failures on Unknown Names/Tags**
**Problem:** When a user specifies `-n unknown-name`, the system silently does nothing if no task or tag matches.

**Why this matters:** Users won't know if they made a typo or if the tag/task doesn't exist. This can lead to confusion and wasted time debugging.

**Suggested fix:** At minimum, log a warning when an input doesn't match any task or tag. Consider whether it should be an error instead.

## Performance Concerns

### 4. **Inefficient O(n²) Tag Lookup**
**Problem:** For each input name that's not a task, you iterate through all tasks and all their tags:
```go
for _, task := range allTasks {
    for _, tag := range task.Tags {
        if tag == name {
```

**Why this matters:** On large playbooks with many tasks and tags, this becomes slow. The performance degradation is quadratic.

**Suggested fix:** Build an inverted index (tag → task names map) once before the main resolution loop. This changes the lookup from O(n*m) to O(1) per tag.

## Missing Components

### 5. **Schema and Documentation Updates**
**Problem:** Missing updates to:
- `schemas/playbook.json` - needs the tags field definition for the Task struct
- After updating README.md, you need to run `make prep-site` to update the site documentation

**Why this matters:** Missing schema updates mean validation tools won't work correctly with the new tags field. The site docs won't reflect the new feature unless you run the make command.

**Suggested fix:**
- Update the JSON schema to include the tags field definition for tasks
- After updating README.md, run `make prep-site` to sync the changes to site docs

### 6. **No Integration Tests**
**Problem:** Only unit tests for `getTaskNamesToRun` exist, but no integration tests showing tags actually working in a playbook.

**Why this matters:** Unit tests verify the resolution logic but don't test the full feature end-to-end. Spot uses test playbooks in `testdata/` directories for integration testing (see examples like `testdata/conf.yml`, `testdata/conf-simple.yml`).

**Suggested fix:**
- Create a test playbook (e.g., `testdata/conf-tags.yml`) that includes tasks with tags
- Add integration tests in `runner_test.go` that use this playbook to verify:
  - Running specific tasks by tag works correctly
  - Multiple tasks with the same tag are all executed
  - Tag resolution works alongside existing task selection

## Code Quality Improvements

### 7. **Empty String Handling**
**Problem:** No validation for empty task names or empty tags.

**Why this matters:** While unlikely, empty strings could cause confusing behavior.

**Suggested fix:** Add validation during playbook parsing or in the resolution logic.

### 8. **Test Improvements**
Your tests use `sort.Strings` before comparison, but this masks the execution order issue. Consider:
- Adding a test that verifies task order is preserved
- Testing edge cases like empty tags, duplicate names, etc.
- Using `assert.Equal` from testify instead of `reflect.DeepEqual`

## Summary

The implementation correctly follows the smart resolution design we discussed, which is great! The main issues are:
1. Moving the logic to where it belongs (PlayBook)
2. Fixing the non-deterministic ordering
3. Improving performance with better data structures
4. Adding user feedback for failed matches
5. Completing documentation and schema updates

Once these are addressed, this will be a solid addition to Spot that maintains consistency with existing patterns.

umputun:
Please let me know if you are still planning to work on this PR. I was hoping to include it in the upcoming release this weekend, but if you are still busy, that is fine - just let me know, and I'll make this change myself.

supermunn:
hey @umputun I was working on it but then was away for a week (had a big move across provinces). I couldn't address everything (integration tests missing, some more tests for edge cases). I just pushed what I have so far, hopefully I'm on the right track. I'd love to continue working on this next week.

umputun:
>  I'd love to continue working on this next week.

Sure. As long as this PR is not abandoned, take your time.

umputun:
The only reason how runner/mocks/playbook.go can get an import cycle is that the Playbook interface (the one you are mocking) has some reference to the runner package. This is not really a mocking issue but rather a design issue. I.e., runner depends on some config structs (like config.Task and config.Destination), but the config package should know nothing about the runner and should not use anything from the runner package.

hope it helps

PR #328, opened 2026-02-09
[Test] Add DOCKER_HOST invariance

zonescape:
There are some tests assuming SSH containers are available on the localhost. All of them will fail if you are using a remote Docker daemon.

### DIFF
--- a/pkg/executor/remote_test.go
+++ b/pkg/executor/remote_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"slices"
+	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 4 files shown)


umputun:
thx for the PR. local docker is a requirement for running integration tests here, so i don't think we need to add this indirection for the remote `DOCKER_HOST` scenario.

also, there's a bug — `conf-line.yml` now checks `host={TESTING_HOST_PORT}` which resolves to `host:port`, but the `line` command uses `SPOT_REMOTE_ADDR` which is host-only, so the grep would fail.

closing this one, but appreciate the effort.

PR #345, opened 2026-05-13
changes to add an SSM remote option

xeoneox:


umputun (changes requested):
this isn't what we discussed in #343. You committed to Session Manager there, with a new connector returning `executor.Interface`. This ships Run Command instead, which is a different model: separate AWS API, base64-over-shell file transfer, polling. The structural piece flagged in #343 (`runner.Connector.Connect` returning the concrete `*executor.Remote`) is also unaddressed - this adds a parallel `SSMConnector` next to it. Two interfaces where one would do, and the lock-in compounds with each future backend.

separate from the direction, here's what I found in the Run Command implementation as shipped. I'm listing these as context, not as a punch-list to fix on this PR. The volume and shape of the problems is part of why I think this needs a rebuild rather than rework.

**SSM correctness**

1. `pkg/executor/ssm.go:99-154` polling loop spins forever when `GetCommandInvocation` returns errors. Every error path is `continue` with no cancellation check or error classification, so once `pollCtx` expires the SDK keeps returning context errors and the loop never exits. Non-retriable errors (`AccessDenied`, `InvalidInstanceId`, throttling) also spin until timeout.

2. `pkg/executor/ssm.go:198-211` Download uses `GetCommandInvocation.StandardOutputContent` as the file transport, but AWS caps that at 24,000 chars unless S3 output is configured. Truncation on a 4-byte boundary passes `base64.DecodeString` without error and Spot writes a corrupted partial file. Plus `len(data) == 0` rejects legitimate zero-byte files and the test asserts that bug (`ssm_test.go:488`).

3. `pkg/executor/ssm.go:204-208` Download also breaks for any file over ~57 bytes: server-side `base64` line-wraps at 76 chars by default, `Run` strips empty lines and joins on `\n`, then `base64.StdEncoding.DecodeString` rejects embedded `\n`. The existing test passes only because the fixture `aGVsbG8gd29ybGQ=` fits on one line. Use `base64 -w 0` server-side or strip whitespace client-side.

4. `pkg/executor/ssm.go:167-168` `maxCommandSize = 16 * 1024 * 1024` is fictional. SSM Run Command has a documented per-parameter limit around 10KB, and the agent-side `ARG_MAX` (~128KB-2MB) bites before that. Files of more than a few KB never reach the agent. The error message even tells the user "use S3 instead" but there's no S3 path.

5. `pkg/executor/ssm.go:180-181` Upload puts the entire base64-encoded file into a shell command, and `Run` at `pkg/executor/ssm.go:76` writes that command via `s.logs.Out.Write(...)` plus `log.Printf("[DEBUG] run SSM command on %s: %s", ...)`. Any copied secret, config file, or binary blob ends up in spot's logs. New disclosure path vs SFTP.

6. `pkg/executor/ssm.go:86-90` `*sendResp.Command.CommandId` deref with no nil check. A throttled or malformed SSM response with `Command == nil` panics the host goroutine.

7. `pkg/executor/ssm.go:174,180,198,266,295,297` shell interpolation uses unquoted `%s` for paths: `mkdir -p`, `echo`, `base64`, `find`, `rm -rf`, `rm`. Any path with a space or shell metacharacter breaks, and paths can be templated from playbook input. Compare to `remote.go` and `local.go` which avoid this via SFTP / `exec.Command`.

8. `pkg/executor/ssm.go:226-247` `SSM.Sync` silently ignores `opts.Exclude`; `local.go` honors it. Silent feature regression for SSM users.

9. `pkg/executor/ssm.go:302` `Delete` parses an English error string (`"No such file"`) for not-exists detection. Locale-dependent.

10. `cmd/spot/main.go:486-496` the SSM connector builds `aws.Config` manually with a static credentials provider. No `config.LoadDefaultConfig`, no session token support, no profile/SSO/instance-role resolution, no default region resolution. Most real AWS setups won't work.

11. `pkg/runner/runner.go:198-223` routing has three silent edge cases:
    - `--ssm` set + host without `ssm_id` falls through to SSH with no warning
    - `--ssm` set + `--local` set, ssm is silently ignored
    - `p.SSM == true` + non-empty `SSMId` + nil `SSMConnector` panics on nil deref

12. `pkg/runner/runner.go:57` `SSMConnector.Connect(ctx, instanceID, hostName, user)` declares a `user` parameter that the only implementation discards with `_`. SSM has no user concept; drop the param.

13. `pkg/executor/ssm_test.go:20-31` hand-rolled `mockSSMClient` violates the project's moq convention (CLAUDE.md says always use moq).

14. `pkg/executor/ssm.go:65,116,126,162,249` bare `//nolint` directives without rule name or justification. Either specify the rule or address the actual lint.

**Unrelated changes that need to come out**

I want this PR scoped to SSM. The following are unrelated and either silently regress behavior or need their own PR.

1. `pkg/executor/local.go:31` removes `$SHELL` env-var detection, hardcodes `/bin/sh`. This affects every local-mode user, not just SSM. Revert.

2. `pkg/runner/commands.go:790-792` `SPOT_ERROR` quote handling: the comment says "Escape single quotes" but `strings.ReplaceAll(tm.err.Error(), "'", "")` deletes them. Silent corruption of error text in on-error scripts. If shell-quoting in on-error is broken, fix it properly (escape with `'\''` or pass via env) in a separate PR.

3. `pkg/runner/commands.go:750` + `pkg/runner/runner.go:429` `templater.secrets` is populated at exactly 1 of 11 call sites. Secret substitution silently works only inside on-error commands and nowhere else. Either complete the plumbing or remove the field.

4. `pkg/runner/runner.go:387` `execErr := &execCmdErr{}` to `var execErr *execCmdErr` is a cosmetic change to an error-handling path. Drop it.

5. `pkg/executor/remote.go:266` `slices.Backward` migration: cosmetic, irrelevant to SSM. If you want to migrate, do it consistently (`local.go:249` still has the manual loop) and in a separate PR.

6. Bare `//nolint` directives in `pkg/config/deepcopy/deepcopy.go:58`, `pkg/executor/local.go:249`, `pkg/runner/runner.go:111`. Bare nolint disables every linter on that line with no audit trail. Specify the rule (`//nolint:gosec // reason`) or address the actual lint.

7. gofumpt rewrites in `pkg/executor/logger.go`, `pkg/runner/commands.go` `Mcopy`, `pkg/config/playbook.go`, the secrets test files, and `pkg/runner/mocks/*.go`. The mocks rewrite will revert on the next `go generate ./...`. Pure noise.

8. `.zed/settings.json` (empty `{}`). Editor config, doesn't belong in the repo.

9. `tmp` empty file at repo root. Stray. Remove.

10. `CLAUDE.md` "Terminal Configuration for MacOS only" section with `/opt/homebrew/...` paths. That's your personal machine setup, not project-wide guidance. Remove.

11. `CLAUDE.md` adding `gofumpt` to project commands. I haven't adopted gofumpt as project policy. If you want to propose that, separate PR with a single chore commit running it everywhere.

**On the direction**

I'd rather see this rebuilt on Session Manager as you proposed in #343. Session Manager tunnels SSH, which means spot's existing `pkg/executor/remote.go` works untouched: no new file-transfer code, no 24KB stdout cap, no fictional 16MB upload cap, no base64-in-logs, no separate vendored SDK service. The connector refactor flagged in #343 (`Connector.Connect` returning `executor.Interface` instead of concrete `*executor.Remote`) is the structural piece that makes future backends like docker/kubectl tractable too.

if there's a reason Run Command is the right call here that I'm missing, I'd want to see that argued out before another round of code. Otherwise let's go back to the Session Manager approach.

xeoneox:
Alright, refactor it is

umputun (changes requested):
strategic piece is still unaddressed.

I asked in round 1 for Session Manager (#343), not Run Command. The refactor reworked the Run Command code but didn't argue why Run Command is the right call. Each of the four reasons I listed for Session Manager still applies: 24KB stdout cap, ~8KB upload cap, base64-in-logs path, and the 127k-line vendored SSM SDK. The connector unification piece (`Connect` returning `executor.Interface`, single `unifiedConnector` instead of two parallel interfaces) matches what I flagged in #343 and I'd keep that shape either way. But the Run Command vs Session Manager question is the one I want answered before another round of code.

**new issues this round**

1. `secrets` binary at repo root is 19.6 MB. In the original commit it was an empty file; the refactor commit added the compiled `spot-secrets` binary. Can't ship in version control. Remove and add to `.gitignore`.

2. `pkg/runner/mocks/connector.go.bak` is committed (2368 bytes). Editor or `moq` rerun artifact. Remove and add `*.bak` to `.gitignore`.

3. `shellQuote` double-quotes. `shellQuote(s)` returns `'...'`, but every call site wraps it again with `'%s'`. For `remote = "foo bar"`, you produce `rm ''foo bar''`, which bash parses as `rm foo bar` (two args, splits on the unquoted space). For paths with single quotes, you get an odd number of quote chars and bash throws a syntax error. Works for simple paths by accident because bash collapses `''foo''` to `foo`. Either drop the outer `'%s'` in callers, or have `shellQuote` not wrap in quotes. Affects `ssm.go:205, 214, 234, 312, 345 (x2), 354, 356`.

**round 1 misses, still standing**

- **SSM #2 Download still capped at 24KB**. `base64 -w 0` only fixes line wrapping; the cap is on `GetCommandInvocation.StandardOutputContent`, which AWS truncates around 24,000 chars unless S3 output is configured. Upload now caps at 8KB with a clear error, Download silently writes a corrupted partial file for anything bigger. The base64 decode happens to succeed on truncated input only when the cut hits a 4-byte boundary, otherwise you get a decode error instead of corrupt bytes. Either add the same guard, or document and check the cap.

- **SSM #5 residual log leak**. `ssm.go:70` has an unconditional `log.Printf("[DEBUG] run SSM command on %s: %s", s.instanceID, cmd)` before the NoLog check at lines 80-83. NoLog only suppresses `s.logs.Out` writes, not the `log.Printf` above it, so base64 payloads from `Upload` still hit logs at `-v`. Move the debug line below the NoLog check, or gate it on NoLog too.

- **SSM #11 partial**. `--ssm` + missing `ssm_id` correctly warns and falls to local now. But `--ssm` + `--local` silently uses SSM if the host has an `SSMId` (the switch hits the SSM case first regardless of `p.Local`). Either error on the conflict, or document the precedence.

- **SSM #12 user param**. `Connect(ctx, hostAddr, hostName, user, ssmID)` still threads `user` through the SSM path which discards it. The unused `_ string` ssmID on the SSH connector at `executor/connector.go:8` is the same shape inverted. Pick one: branch on type at the call site, or have a thinner interface that doesn't carry parameters the other side discards.

- **Scope #5 inconsistency**. `slices.Backward` was applied to `remote.go:266` but not `local.go:249`, which still has the manual reverse loop with a new `//nolint:intrange`. Pick one across the file.

- **Scope #7 cosmetic noise still in local.go**. Blank-line removals at lines 35, 43, 51, 66 are gofumpt-style and unrelated to SSM. Revert.

- **Scope #4 cosmetic error change**. `execErr := &execCmdErr{}` → `var execErr *execCmdErr` at `runner.go:387` is still a cosmetic edit on an unrelated path. Revert.

**new small bugs introduced by the refactor**

- `unifiedConnector.Connect` and `connectSSM` ignore their `ctx` parameter (`main.go:110, 118`) and pass `context.Background()` to AWS config and SSH client setup. Cancellation won't propagate. Use the passed ctx.

- AWS credentials for SSM are reused from `opts.SecretsProvider.Aws.{AccessKey, SecretKey, Region}` at `main.go:64-71`. That couples two unrelated configs: someone using SSM without an AWS secrets provider has no creds path beyond env/IRSA, and someone with SecretsProvider in region A can't target SSM in region B. SSM needs its own group of flags.

**addressed in this round**

- polling loop now classifies retriable vs non-retriable via `isRetriableSSMError`
- `base64 -w 0` on read/write fixes line-wrapping
- nil check on `Command.CommandId`
- `Sync` honors `Exclude`
- `Delete` uses `test -f || test -d`, locale-independent
- `awsconfig.LoadDefaultConfig` so profile/SSO/IRSA work (still no SessionToken in static creds path)
- moq-generated `ssm_mock.go`
- `SPOT_ERROR` now escapes `'\''` correctly instead of deleting quotes
- `templater.secrets` populated at all 11 call sites
- `.zed/settings.json`, `tmp`, CLAUDE.md SSM/gofumpt sections removed
- `$SHELL` env-var detection restored in `local.go`
- bare `//nolint` in `runner.go:111` now has rule + justification
- `Connect` returns `executor.Interface`, single `unifiedConnector` instead of parallel interfaces

still want the direction discussion before another round. If you want to argue for Run Command over Session Manager, that's fine, but I'd want the case made instead of more code on top.

xeoneox:
I'm not tied to RunCommand, it's easiest to test with, and that's why I opted for it initially. I'm thinking later on, if SSM goes well, I still don't want to add the equivalent for Azure or GCP. Anecdotal evidence suggests SSM and Teleport are the most popular solutions. I think your way would lead to a better fit later if I continue trying to expand connection methods. I will attempt to refactor this in a separate push from the other remaining issues.

umputun:
your message reads like agreement but the load-bearing clause is "if I continue trying to expand connection methods" - i.e. you're saying your way is fine for you because you're not planning to expand. That's not the question I was asking.

the Session Manager call isn't about generalizing the connector for future Azure/GCP backends. It's about the RunCommand-specific defects that don't go away with refactoring:

1. 24KB `GetCommandInvocation.StandardOutputContent` cap. Download silently writes corrupt files for anything bigger. Only fix is S3 output or a chunked-transfer protocol, both of which are new code paths you'd be writing on top of what's already here.

2. Upload puts the full base64 payload into a logged shell command. Refactoring doesn't change that; only moving file transfer off `Run` does.

3. AWS creds piggy-backed on `--secrets.aws.*`. SSM can't target a region different from your secrets, can't use a different profile, can't take a session token. Separate config group is the real fix.

these are why I asked for Session Manager: it tunnels SSH and inherits `pkg/executor/remote.go` unchanged - no 24KB cap, no logged payload, no separate AWS surface, no S3 fallback to build. If you want to keep RunCommand, I want the technical case for why these limitations are acceptable, not the "I'm not expanding" framing. Otherwise let's go to Session Manager and skip another round of patches on RunCommand.

xeoneox:
ok, first, the session manager / base64 / output cap refactor. Then, I will push fixes for the remaining issues

PR #346, opened 2026-06-26
feat: allocate pseudo-terminal (TTY) for Line command executions and fix append sudo wrapping

vakechu:
Fix for sudo error in **line** command
```
2026/06/26 11:39:34.640 [WARN]  ! sudo: a terminal is required to read the password; either use ssh's -t option or configure an askpass helper
2026/06/26 11:39:34.641 [WARN]  ! sudo: a password is required
2026/06/26 11:39:34.641 [WARN]  ! failed command "add-line" on host 10.100.100.100:55222 (): can't execute line append on 10.100.100.100:55222: failed to run command on remote server: Process exited with status 1
```


umputun (changes requested):
main blocker is the sudo append path.

`ec.wrapWithSudo(fmt.Sprintf("%s -c %q", ec.shell(), operationCmd))` at commands.go:588 builds `sudo /bin/sh -c "echo '...' | tee -a FILE > /dev/null"`. `%q` is Go double-quote quoting, not shell quoting, so it doesn't escape `$` or backticks, and the single quotes around the content are just literal chars inside the outer double-quoted string. Two problems fall out of that:

1. `$VAR`, backticks and `$(...)` in `append:` now expand in the remote login shell before sudo runs. The old `echo '...' | sudo tee` kept them literal. So appending a value like `root:$6$salt$hash` or a crontab line with `$PATH` gets silently mangled. The non-sudo path (line 586) still single-quotes the content, so the same `line:` command behaves differently with and without sudo.

2. a `'` inside `append:` breaks out of the quoting. That was already possible before, but the old form only elevated `tee`, now the whole pipe runs under `sudo sh -c`, so a breakout runs as root.

`file` is also unquoted (`tee -a %s`), pre-existing but worth fixing here too.

for the fix, escape the content the same way `wrapWithSudo` already escapes the sudo password (`'` -> `'\''`) instead of `%q`, and single-quote `file`. Keep the `sudo -S sh -c` regrouping, that part is right, it's just the quoting that leaks.

a few more:

- `RequestPty: true` is hardcoded on both Run calls (commands.go:584, 606), including the non-sudo grep/sed/tee. Any sshd that refuses a pty (`PermitTTY no`, forced-command keys) now fails every `line` command, not just the sudo ones. Password sudo (`sudo -S`) reads the password from stdin so it doesn't need a tty, only `requiretty` does. Can you gate the pty on `Options.Sudo`?

- the `\r` trim in sshRun (remote.go:349) runs on every remote command's captured output, not just pty runs. The `\r` only shows up because the pty turns on onlcr, so scope it to `opts.RequestPty` rather than the shared path.

- `session.RequestPty("xterm", 80, 40, ...)` (remote.go:320) is `(term, h, w)`, so that's 80 rows by 40 cols. Looks like height and width got swapped, probably meant `40, 80`. No effect on grep/sed/tee since they ignore geometry, but worth flipping.

- no tests for any of the new behavior. The existing append-sudo test uses passwordless sudo with a value that has no shell-special chars, so none of the quoting, the sudo_password path, or the pty path is covered, pls add cases for `$`/backtick/`'` in `append:` plus the sudo_password sudo path.

umputun (changes requested):
still blocked on the quoting from the last round, and the Jul 9 push made it wider.

`pkg/runner/commands.go:619` - `sh -c "echo '%s' | tee -a %s > /dev/null"` is parsed by the remote login shell before sudo runs, so `$VAR`, backticks and `$(...)` in `append:` expand and a `'` breaks out and runs as root. Swapping `%q` for hand-written double quotes changes nothing there. What is new: the wrapper is now on the non-sudo path too, so `append: "PATH=$HOME/bin"` was written literally before and now expands. Fix: keep the program static and pass the data as arguments, `sh -c 'printf "%s\n" "$1" >> "$2"' sh '<content>' '<file>'`, with `'` in content and file escaped as `'\''` the same way `wrapWithSudo` escapes the password. That form goes through the existing `wrapWithSudo` as is, leaves stdin free for the `sudo -S` password pipe, and works without sudo. Pls single-quote `file` in the grep check at 615 as well. Tests are still missing: `append:` with `$`, backtick and `'` in the value, plus append with `sudo_password`, in `Test_execLine`.

housekeeping: the title still describes PTY allocation that is no longer in the diff, `pkg/executor/executor.go:26` and `pkg/executor/remote.go:362` have leftover blank lines from the removal, and the branch conflicts with master after the `discardOutput` change in `Line()`.

the pty, `\r` and geometry points from the last round are moot now that the code is gone.

PR #349, opened 2026-07-03
feat: add template command (Ansible-style template rendering + copy)

robertobarreda:
Adds a `template` command that renders a local Go text/template file with env vars, registered variables, loaded secrets and `SPOT_*` variables, then uploads the result to a remote host. The upload reuses the existing copy-push path so `mkdir`, `force`, `chmod+x`, `mode` and `sudo` work the same as `copy`.

New config struct `TemplateInternal{Source, Dest, Mkdir, Force, ChmodX, Mode}` with `execCmd.Template()` — full test coverage covering basic render, env, secrets, sudo, mkdir, chmod+x, mode, cond skip, missingkey error, idempotent re-upload, single-quoted export propagation, and error paths.

Bugs fixed along the way: single-quoted `__SQ__:` marker leaking into template data, missing `SPOT_ERROR` in template vars, undefined keys silently rendering empty (now `missingkey=error`), and pointer-vs-value copy bug in `Mcopy`/`Msync`/`Mdelete`.

Includes schema update (`templateSpec` in `playbook.json`), README and llms.txt docs.

```yaml
- name: render nginx config
  template: {src: "templates/nginx.conf.tmpl", dst: "/etc/nginx/nginx.conf", mkdir: true}
  env:
    APP_ENV: production
    APP_PORT: "8080"

- name: render with a loaded secret
  template: {"src": "templates/db.conf.tmpl", "dst": "/etc/app/db.conf", "mkdir": true, "mode": "0644"}
  options:
    secrets: [db_password]
```


umputun (changes requested):
neat idea, and it opens up some nice possibilities, per-host config with conditionals, generating init scripts, dropping a secret straight into a config file in one step. Worth saying up front that the simpler cases don't really need it, plain per-host content or a single value drop already works with a `script` heredoc (which does SPOT_*/env/secret substitution too), so template mainly earns its place on the Go-template logic and the one-step secret-into-config. Good direction overall, a few things to sort before merge, one of them a blocker.

**1. not idempotent (blocker).** with `force: false` the rendered file gets re-uploaded and overwritten on every run. `copyPush` skips on size+mtime+mode, but the temp file from `os.CreateTemp` gets a fresh mtime each run, so it never matches the remote and always re-uploads. For a config tool that churns mtime and can trigger service reloads/restarts. Compare rendered content against the remote (checksum), or set the temp file's mtime deterministically from the rendered bytes so an unchanged render actually skips.

**2. single-quoted registered vars render the `__SQ__:` marker.** `maps.Copy(tplData, ec.cmd.Environment)` at commands.go:657 copies env values raw, but a var set via a single-quoted export/setvar is stored as `__SQ__:value`, and only `templater.apply` strips that prefix (commands.go:890-900). So `{{ .FOO }}` for such a var renders the literal `__SQ__:bar`.

**3. `SPOT_ERROR` is missing from the template vars.** the SPOT_* map is hand-built at commands.go:643-656 and omits `SPOT_ERROR`, which `templater.apply` exposes. Both this and #2 disappear if `Template()` reuses templater's var-map + marker-strip logic instead of rebuilding it, so extracting that into a shared helper is probably the cleanest fix for both.

**4. rendered files always land 0600, add a `mode` field.** the temp file is 0600 and `Upload` chmods the remote to the source mode, so every rendered file is owner-only regardless of `dst`. A service running as another user can't read it. Add a `mode:` option, default 0600 so secret-bearing renders stay locked down and users set 0644 for plain configs. Don't preserve the .tmpl's own mode, a 0644 template would make a secret render world-readable.

**5. `missingkey=zero` hides typos.** an undefined key like `{{ .APP_PROT }}` renders empty with no error, so a config can silently ship `listen ;` or an empty secret. Switch to `Option("missingkey=error")` so a typo fails the command instead of writing a blank value.

smaller stuff:

- `resp.details` shows the temp path as the source: `{template: /tmp/spot-templateNNN -> /etc/...}`. Rebuild it with the real `src` so the output means something.
- `ecCopy.cmd.Copy` sets `Source`/`Dest`/`Direction`, but `copyPush` takes the paths from its args and never reads `Direction`. Drop the three dead assignments.
- docs: README uses `testdata/nginx.conf.tmpl` as the example src while llms.txt uses `templates/`, pick `templates/` in both (`testdata/` is a test-fixture convention). The "respects `cond` like every other command" line isn't right (copy/sync/delete/wait don't support cond), just say it supports `cond`. And `validate`'s godoc list still omits `template`.
- tests: the register-vars-in-templates path is only covered by injecting `Environment` directly. Add one that drives a `register` script through the runner into a template, with a single-quoted export so it catches #2. There's also no force=false/identical-content case.

tests and lint are green, and the secret handling checks out (rendered to a 0600 temp, cleaned up, never logged, no injection). No need to rework the copyPush reuse itself, the work is in the idempotence fix and the two var-map bugs.

robertobarreda:
> neat idea, and it opens up some nice possibilities, per-host config with conditionals, generating init scripts, dropping a secret straight into a config file in one step. Worth saying up front that the simpler cases don't really need it, plain per-host content or a single value drop already works with a `script` heredoc (which does SPOT_*/env/secret substitution too), so template mainly earns its place on the Go-template logic and the one-step secret-into-config. Good direction overall, a few things to sort before merge, one of them a blocker.
> 
> **1. not idempotent (blocker).** with `force: false` the rendered file gets re-uploaded and overwritten on every run. `copyPush` skips on size+mtime+mode, but the temp file from `os.CreateTemp` gets a fresh mtime each run, so it never matches the remote and always re-uploads. For a config tool that churns mtime and can trigger service reloads/restarts. Compare rendered content against the remote (checksum), or set the temp file's mtime deterministically from the rendered bytes so an unchanged render actually skips.
> 
> **2. single-quoted registered vars render the `__SQ__:` marker.** `maps.Copy(tplData, ec.cmd.Environment)` at commands.go:657 copies env values raw, but a var set via a single-quoted export/setvar is stored as `__SQ__:value`, and only `templater.apply` strips that prefix (commands.go:890-900). So `{{ .FOO }}` for such a var renders the literal `__SQ__:bar`.
> 
> **3. `SPOT_ERROR` is missing from the template vars.** the SPOT_* map is hand-built at commands.go:643-656 and omits `SPOT_ERROR`, which `templater.apply` exposes. Both this and #2 disappear if `Template()` reuses templater's var-map + marker-strip logic instead of rebuilding it, so extracting that into a shared helper is probably the cleanest fix for both.
> 
> **4. rendered files always land 0600, add a `mode` field.** the temp file is 0600 and `Upload` chmods the remote to the source mode, so every rendered file is owner-only regardless of `dst`. A service running as another user can't read it. Add a `mode:` option, default 0600 so secret-bearing renders stay locked down and users set 0644 for plain configs. Don't preserve the .tmpl's own mode, a 0644 template would make a secret render world-readable.
> 
> **5. `missingkey=zero` hides typos.** an undefined key like `{{ .APP_PROT }}` renders empty with no error, so a config can silently ship `listen ;` or an empty secret. Switch to `Option("missingkey=error")` so a typo fails the command instead of writing a blank value.
> 
> smaller stuff:
> 
> * `resp.details` shows the temp path as the source: `{template: /tmp/spot-templateNNN -> /etc/...}`. Rebuild it with the real `src` so the output means something.
> * `ecCopy.cmd.Copy` sets `Source`/`Dest`/`Direction`, but `copyPush` takes the paths from its args and never reads `Direction`. Drop the three dead assignments.
> * docs: README uses `testdata/nginx.conf.tmpl` as the example src while llms.txt uses `templates/`, pick `templates/` in both (`testdata/` is a test-fixture convention). The "respects `cond` like every other command" line isn't right (copy/sync/delete/wait don't support cond), just say it supports `cond`. And `validate`'s godoc list still omits `template`.
> * tests: the register-vars-in-templates path is only covered by injecting `Environment` directly. Add one that drives a `register` script through the runner into a template, with a single-quoted export so it catches #2. There's also no force=false/identical-content case.
> 
> tests and lint are green, and the secret handling checks out (rendered to a 0600 temp, cleaned up, never logged, no injection). No need to rework the copyPush reuse itself, the work is in the idempotence fix and the two var-map bugs.

Thanks for the detailed review. I tried to address all your concerns.

umputun (changes requested):
most of the list is done. Items 2, 3 and 5 are fixed and the docs corrections are in. Four things still block, and two of them come out of the fixes themselves rather than the original feature.

**1. still not idempotent (blocker, item 1).** the content-derived mtime can't round-trip. `commands.go:721` builds the time from 63 bits of the hash, so it's a Unix second anywhere up to ~9.2e18, but `pkg/sftp` sends mtime as a uint32: `attrs := times{uint32(atime.Unix()), uint32(mtime.Unix())}` in `vendor/github.com/pkg/sftp/client.go:608`. Only hashes that fall inside the uint32 range survive, which is a vanishing fraction of them. Common filesystems can't represent most of that range either.

two other paths still force an upload regardless of what the hash produces:

- sudo: the upload goes to `ec.uniqueTmp(...)`, a fresh random path each run, so `sftpClient.Stat` on it misses and the `isSame` block at `remote.go:463` never runs. The real `dst` is never compared at all, it just gets `mv -f`d over.
- chmod+x: `isSame` compares mode, and the remote ends up with execute bits while the staging file stays 0600, so it can't match on any later run.

comparing the rendered content and the wanted destination mode against `dst` itself would cover all three.

**2. the shared var-map changed substitution for every command (blocker).** `templater.vars()` is the right idea, but `apply` now iterates it:

```go
for k, v := range tm.vars() {
```

map iteration order is unspecified and varies between runs. The old code applied the `SPOT_*` set in a fixed sequence and then env. Substitution is sequential text replacement, so the order decides the result whenever one value contains another variable's token. With `env: {CFG: "{SPOT_TASK}.conf"}` and `dst: "/etc/{CFG}"`, repeated identical evaluations give both `/etc/deploy.conf` and the unexpanded `/etc/{SPOT_TASK}.conf`. Same playbook, different result, and `apply` is shared, so this reaches copy, sync, delete, script, wait, echo and line, not just template.

the merged map also lets an env key overwrite a `SPOT_*` built-in, which it couldn't do before.

keep `vars()` for the template data, but put back the explicit built-ins pass followed by the env pass in `apply` so existing behavior stays as it was.

**3. `mode` is applied to the local staging file (blocker, item 4).** I meant `mode` as the destination mode. The staging file needs to stay 0600. `os.Chmod(tmpName, ...)` at `commands.go:714` can widen the temp file that already holds the rendered secrets, and `sftpUpload` then takes the remote permissions from that local stat (`remote.go:513`):

- on a runner with a shared temp dir, `mode: "0644"` plus `options.secrets` makes the rendered secret world-readable there for the length of the upload.
- on Windows the default `0600` doesn't hold either. `os.Chmod` only toggles the readonly attribute, and `os.Stat` reports any writable regular file as 0666 (`os/types_windows.go`), so 0666 is what gets applied on the remote. Windows is a released target in `.goreleaser.yml`.

these need to be two separate things: the staging mode, always 0600, and the wanted destination mode, compared and applied against `dst` on its own. Folding the second into the first is also what keeps item 1 broken, since a 0600 local file never matches a 0644 remote.

**4. `chmod+x` is silently dropped when `mode` is set.** `chmodX := ec.cmd.Template.ChmodX && ec.cmd.Template.Mode == ""` at `commands.go:729`. `{mode: "0644", chmod+x: true}` gives a non-executable file, reports success, and logs nothing. Config and validation accept both fields and the docs present them as independent options. Either fold the execute bits into the final mode or reject the combination with an error.

smaller stuff:

- `commands.go:693`: `defer tmp.Close()` is registered before the remove defer, so LIFO runs the remove while the handle is still open. Windows opens files without `FILE_SHARE_DELETE`, so the remove fails and the rendered secret file stays in temp, with no retry after the close. Close before the upload, or use one deferred func that closes and then removes.
- the nginx example can't run as written. `README.md:430` sets no `env`, but `templates/nginx.conf.tmpl` at `README.md:448` uses `{{ .APP_PORT }}` and `{{ .APP_ENV }}`, and both are absent, so execution aborts on the first one. Same in `site/docs-src/index.md` and `site/docs/llms.txt`. Add the `env:` block to the example or drop those keys from the template body.
- `{{ range .List }}` in that same block can't work either. Template data is `map[string]string`, so `.List` is either absent, which errors, or a string, and execution fails because text/template can't range a string.

on my earlier points that are still open:

- the dead `Source`/`Dest`/`Direction` assignments are still at `commands.go:732`. `010c06a2`'s message says it dropped them but the commit doesn't touch those lines.
- both tests I asked for are there, but neither pins what it's named for. `template force=false duplicates are skipped` asserts the details prefix and the file content, and both hold whether the second upload was skipped or not, so it needs an assertion that tells a skip from a re-upload, which final content and mtime don't. The register test hand-injects `Environment` rather than driving a `register` script and then a template through `Process`, so the dispatch arm in `runner.go` is still not executed by any test and no fixture has a `template:` key. A playbook fixture running register then template would cover both.

the feature shape is good and I don't want it reworked. The variable and missingkey fixes are in good shape. The work left is the idempotence approach, putting `apply`'s ordering back, and separating staging mode from destination mode, which also settles the chmod+x interaction.

robertobarreda:
All four blockers and the smaller items are addressed in e0153ff6.

**1. idempotence (blocker)** — dropped the content-derived mtime approach entirely. `Template` now compares the rendered content (`sha256sum`) and the wanted mode (`stat -c %a`) against `dst` itself before uploading, wrapped with sudo when set. Any mismatch (including sudo and chmod+x paths) uploads and applies the mode; a full match skips with `skip: identical` in the details. A missing/unreadable remote file counts as a mismatch, so it cannot skip wrongly.

**2. apply ordering (blocker)** — reverted `apply` to the explicit built-ins pass (fixed sequence) followed by the env pass. Deterministic across runs, and an env key can no longer shadow a `SPOT_*` built-in. `vars()` stays as the template-data builder. Added tests for both properties.

**3. staging mode (blocker)** — staging file stays at `os.CreateTemp`'s 0600. The wanted destination mode is a separate value (explicit `mode` or 0600 default), applied to `dst` with a chmod after upload. Because the chmod always runs, the Windows-client case gets exact perms on the remote too.

**4. chmod+x + mode (blocker)** — execute bits are folded into the final mode (`mode |= 0111`), so `{mode: "0644", chmod+x: true}` is 0755 and no combination is silently dropped. Test pins 0755.

**smaller items**
- staging file is closed before the remove defer runs (no LIFO open-handle remove).
- dead `Source`/`Dest`/`Direction` assignments removed; details built from the real `src`.
- nginx example gets the missing `env:` block (`APP_PORT`, `APP_ENV`) in README, docs-src and llms.txt; `{{ range .List }}` removed from the syntax line since the data is `map[string]string`.
- force=false test now asserts `skip: identical` (distinguishes skip from re-upload); added cases for changed-content re-upload, chmod+x idempotence, and mode-only re-apply.
- new playbook fixture (`register_into_template.yml` + `template_registered.tmpl`) drives a single-quoted `register` script then a `template` through `Process`, covering the runner dispatch arm and the `__SQ__` marker strip (renders `value=sq-$value`).

Full suite (`go test ./...`), race on the runner package, and golangci-lint (CI's go1.25 + v2.11.1) are green.

umputun (changes requested):
all four blockers are fixed, including the sudo and chmod+x paths on the idempotence one. `apply`'s ordering is back to the explicit pre-PR sequence, and the doc and test items from both rounds are in, the force=false test now tells a skip from a re-upload and the register fixture drives the runner dispatch arm for real. One thing left that blocks, plus a small one.

**1. the idempotence check only works on GNU userlands (blocker).** `sha256sum` and `stat -c %a` at `commands.go:762,771` don't exist on macOS or the BSDs, which have `shasum -a 256` and `stat -f %Lp`. Both probes error, `templateMatchesRemote` returns false, and it falls through to the upload. That path can't skip either: `Local.Upload` compares size, mtime and mode (`local.go:131-139`) and the staging file from `os.CreateTemp` gets a fresh mtime each run, so `isSame()` never holds and unchanged content rewrites `dst` every time. The `--local` flag is documented at README:149 and `options: {local: true}` at README:941, and darwin and freebsd are both in `.goreleaser.yml`, so this is the same mtime churn the last two rounds blocked on, on platforms spot ships.

build the alternation first and wrap it once through a shell, rather than wrapping each command on its own, or `sudo` covers only the first branch and the fallback fails again on a root-owned file:

```go
ec.wrapWithSudo(ec.shell() + " -c " + shellQuote("sha256sum "+q+" 2>/dev/null || shasum -a 256 "+q))
```

same shape for `stat -c %a || stat -f %Lp`. If the probe emits a tagged line the code scans for, that also settles the other thing here: `commands.go:763,779` reads `out[0]`, the only such index in the package, so a host whose shell rc writes to stdout shifts the hash onto `out[1]` and the skip silently stops firing.

**2. a loaded secret can overwrite a `SPOT_*` built-in.** `vars()` guards the built-ins against env keys at `commands.go:944-946`, then `Template` overlays secrets onto the same map at `:672-675` with no guard, so a secret named `SPOT_TASK` replaces the built-in and the rendered file gets the secret where the task name should be. Mirror the `vars()` guard when merging the secrets, and add a collision case next to `Test_templaterVars`.

umputun (changes requested):
both items from the last round are in, the probe is wrapped once with a tagged line and a secret can't shadow a built-in any more. One thing still blocks, and it comes out of that probe fix, plus a few small ones.

**1. the wrapped probe breaks on `--local` (blocker).** `template.go:189` builds `/bin/sh -c '<probe>'`, and `Local.Run` at `pkg/executor/local.go:40-45` strips a matching `<shell> -c '...'` wrapper with a bare quote trim. The probe's own `'\''` escapes are left unbalanced, `/bin/sh` fails with `unexpected EOF while looking for matching '`, the probe returns false and every run re-uploads and re-chmods an unchanged file, with the syntax error on the WARN log each time. It fires whenever `$SHELL` is unset or `/bin/sh` (cron, systemd, docker, CI), and always when `--shell` is set together with `options: {local: true}`. With `$SHELL=/bin/zsh` it happens to work, which is why it passes interactively.

keep the wrapper, the raw probe doesn't parse under csh or fish login shells over ssh, and prefix it with `exec` on the non-sudo path so the strip can't match:

```go
c := "exec " + ec.shell() + " -c " + shellQuote(probe)
if ec.cmd.Options.Sudo {
    c = ec.wrapWithSudo(ec.shell() + " -c " + shellQuote(probe))
}
```

add a test that drives `Template` twice through the `Local` executor with `SHELL=/bin/sh` and checks `skip: identical` on the second run plus an unchanged mtime. Mocking the probe won't catch this one.

**smaller stuff, all in the same probe and docs:**

- `template.go:172`: BSD `stat -f %Lp` prints only the low three digits, so `mode: "4755"` or a sticky bit never matches on a macOS or FreeBSD remote and re-uploads every run. `%Mp%Lp` gives the full value and still parses for plain modes. The "bsd mode spelling" subtest doesn't exercise the fallback, both spellings sit in one string and the `stat -c` arm matches either way.
- same line: `stat` without `-L` reports the link, while `sha256sum`, the chmod and the upload follow it. A symlinked `dst` never matches on mode, and when the requested mode equals the link's own (0755 on macOS, 0777 on linux) it reports `skip: identical` and leaves the target's mode wrong. `stat -L` on both branches, busybox and BSD support it.
- README:523 (and docs-src, llms.txt:328): "same chmod+x semantics as copy" isn't right. `copy` adds `+x` on the source mode so a 0644 script ends up 0755, `template` adds it to the 0600 default and ends up 0711. The executable-script example at README:499 sets no `mode`, so `run.sh` is unreadable for anyone but the owner. Say chmod+x adds the execute bits to `mode` and put `mode: "0644"` in that example. The 0600 default stays.
- README:490 (and docs-src, llms.txt:297): nothing says a missing key is an error rather than empty, and the syntax line shows `{{ if .X }}`, which also fails when `X` is absent under `missingkey=error`. One line that an unknown key fails the command, and that `{{ if index . "X" }}` is the way to make a key optional.

umputun (commented):
round four's items are all in. The `exec` prefix at `template.go:191` defeats `Local.Run`'s wrapper strip, and `Test_execTemplateLocalProbeSkipsIdentical` pins it with a real `Local` executor under `SHELL=/bin/sh`, checking both `skip: identical` and an unchanged mtime. `stat -L -f %Mp%Lp` gives the full mode on macOS (`0644`, `4755`, `1777`), and `ParseUint` puts the GNU and BSD spellings on the same value. Both probe branches carry `-L`. The chmod+x and missing-key wording is right on all three doc surfaces.

`ce777db6` settles something I was about to raise. The old guard skipped a selected secret whenever the key was already in `tplData`, and `vars()` writes env keys there even when the value is `""`, so `env: {DB_PASSWORD: ""}` together with `options: {secrets: [DB_PASSWORD]}` rendered `password=` and `missingkey=error` could not fire, because the key existed. `slices.Contains(spotVarNames, k)` is the right narrowing.

one question on that fix rather than an objection. With `v != ""`, a secret that loads as an empty value now stays out of `tplData` altogether, so `{{ .DB_PASSWORD }}` fails with "map has no entry for key" even though the secret was requested and loaded fine. Is that what you want, or should an empty loaded secret render empty?

two small things left, neither blocking.

**1. details string, `template.go:61-66`.** The brace closes before the suffixes are appended, so a sudo template prints `{template: a -> b}, sudo: true, chmod: +x` while every other command keeps them inside, e.g. `{copy: %s -> %s, sudo: true, chmod: +x}` at `commands.go:282`. The skip path at `template.go:48` in the same file gets it right. The tests assert with `Contains` on `"sudo: true"` and `"chmod: +x"`, so nothing pins the brace. Building the suffix first and formatting once covers it.

**2. the nginx example, `README.md:493`,** and the same block in `site/docs-src/index.md` and `site/docs/llms.txt`. The template body has a bare top-level `server { }` and the playbook writes it to `/etc/nginx/nginx.conf`. That block has to sit inside `http { }`, and the file also needs an `events { }` section, so nginx rejects it at parse time. Spot exits 0 either way, so it turns up at the next reload rather than in the run. `worker_processes {{ .SPOT_REMOTE_PORT }}` also renders `worker_processes 22` on a default-port host. Pointing `dst` at a drop-in like `/etc/nginx/conf.d/app.conf` and dropping the `worker_processes` line handles both.

umputun (changes requested):
round five's items are in. One thing blocks that I should have caught earlier, plus two small ones.

**1. `--dry` fails on a template that reads a registered variable (blocker).** `renderTemplate` at `template.go:89` sets `missingkey=error` unconditionally, but under `--dry` no script actually runs. `Dry.Run` echoes the command text instead of executing it, so `isSetVarLine` never matches and `register`/`setvar` values never reach `cmd.Environment`. The render then fails the command, and `runTaskOnHost` returns on the first error, so every later command and every later task is skipped. `copy` and `script` leave the same unresolved variable as a literal and exit 0, since `templater.apply` only substitutes keys it has.

it fires on the pattern README:487 documents and `testdata/register_into_template.yml` exercises:

```
real: completed command "render" {template: app.conf.tmpl -> /tmp/x.conf}   x.conf = "version=1.2.3"
dry:  failed command "render": can't execute template "app.conf.tmpl": at <.APPVER>: map has no entry for key "APPVER"
```

changing `missingkey` doesn't cover it. With `missingkey=default`, `{{ len .APPVER }}` still errors; with `zero`, `{{ slice .APPVER 0 1 }}` does. Under `--dry` the template shouldn't execute at all: parse the source, resolve and validate `mode`, report the operation it would perform, and keep `missingkey=error` on the real path. `execCmd` has no dry flag today and `Process.Dry` is read only in `pickCmdExecutor` at `runner.go:349`, so the signal has to reach `Template()` first. A test needs register, then template, then a trailing command under `--dry`, with a value-dependent expression in the template body so a missingkey change alone can't pass it.

smaller stuff:

- `command.go:463`: the numeric-mode rejection prints the decoded integer, so `mode: 0644` gives `template mode must be a quoted octal string, got 420`. YAML resolves `0644` as legacy octal and 420 appears nowhere in the playbook. Drop the number and show the wanted form instead. `%#o` isn't the fix either, `mode: 644` would print `01204`.
- `template.go:114`: `templateMode` returns an `os.FileMode` holding a raw octal. It stores fine, but it doesn't carry Go's `ModeSetuid`, and `applyTemplateMode`'s comment says the full mode value is used so setuid/setgid/sticky survive. That's true through `%o` and false through `Perm()`, so anyone acting on the comment and switching to `os.Chmod` on the `--local` path drops the bits with nothing failing. The only local template test uses 0600, so it wouldn't show. Returning `uint32` makes the comment true.

umputun (changes requested):
I can review one more revision of this. If the two blockers below are still there, I'll close the PR. `mode` and the remote probe were both things I asked for in round one, and six rounds on I'd rather take them back than keep iterating on them.

two things block. Both are on the non-sudo path, running as an ordinary user who owns the file.

**1. a template installed with a non-writable mode can't be updated.** `template.go:66-70` runs `uploadRenderedTemplate` before `applyTemplateMode`, and both `Local.Upload` (`os.Create` at `local.go:282`) and `sftpUpload` (`sftpClient.Create` at `remote.go:470`) open the destination with `O_RDWR|O_CREATE|O_TRUNC`. So `mode: "0400"` or `"0444"` installs fine and skips while unchanged, then the first run with changed content fails with permission denied and leaves the old bytes in place. Someone has to chmod or replace the file on the host before the next run works. That's secret rotation, and 0400 is the mode you'd pick for it. Running with `sudo` avoids this particular failure, `copyPush` replaces the inode with `mv -f` there.

reproduced on the local executor at `5def377a`. The SSH path I traced through `sftpClient.Create` rather than ran, so treat that half as read, not tested.

**2. a YAML anchor or merge key gets an unquoted numeric mode past the guard.** `validateTemplateModeYAML` at `command.go:457-470` walks the direct entries of the mapping and returns early on an alias, so `template: {<<: &defaults {mode: 0644}, src: x, dst: y}` loads. The later decode resolves the merge and yields `Mode: "420"`, which `templateMode` parses as octal 0420.

both spellings miss what the playbook asked for, in opposite directions:

- `0644` becomes 0420, owner read-only. Unchanged reruns skip, and a changed render hits blocker 1.
- `0400` becomes 0256, owner write-only with group `r-x` and other `rw-`. The first install hands other users read and write on a rendered secret, and later runs fail for the same reason as blocker 1.

the guard needs to validate the mode that actually takes effect, merges and aliases included. Rejecting every integer found while walking isn't right either, a numeric inherited value overridden by a quoted direct one should still load.

**on the scope.** what I'd rather have is `template` as render plus copy: no `mode` option, no `sha256sum`/`stat` probe. That's my round-one ask I'm reversing, not something you got wrong. It costs the skip on an unchanged destination, so every run re-uploads, and it still needs the rendered file to land with safe permissions rather than whatever the staging file happens to carry.

that's a preference, not a condition. Fixing the two blockers on the current scope is fine by me too. Either way those two are the bar and nothing below them is.

smaller stuff, optional:

- `command.go` checks the YAML type of `mode` but never its value, so `mode: "0999"` loads, connects, runs everything ahead of the template command and then aborts per host mid-task. `Cmd.validate` is the place for a value check rather than the YAML hook, since TOML doesn't go through the hook. And the `mode` row at `README.md:481` (plus `site/docs-src/index.md` and `site/docs/llms.txt:289`) could say "quoted octal string".

robertobarreda:
I'm sorry. I'll close the PR. I appreciate your time and effort on each iteration. Thank you.

I don't have the time or the focus this task needs right now.

I started this task several weeks ago and underestimated its scope and complexity.
My goal was to contribute to this project and bring it closer to Ansible's core features.

PR #350, opened 2026-07-03
Bump golang.org/x/crypto and golang.org/x/net

paskal:
Bumps `golang.org/x/crypto` v0.47.0 → v0.53.0 and `golang.org/x/net` v0.49.0 → v0.56.0 (vendored). govulncheck reported 9 vulnerabilities on live call paths, including an SSH panic (GO-2026-5013) and an ssh-agent forwarding constraint bypass (GO-2026-5006) via the connector, and HTTP/2 issues (GO-2026-4918, GO-2026-5026) via inventory loading over HTTP; it is clean after the bump.

The new x/crypto `agent.NewClient` starts a background reader goroutine on the agent connection, so the previously benign unclosed `SSH_AUTH_SOCK` connections would now leak a goroutine in addition to the socket on every agent-authenticated connection. The auth-time agent connection is now closed right after the SSH handshake, and the forwarding connection is closed when the SSH client terminates; a failed forwarding setup also closes the client instead of leaking the connection.

Split out of the original combined PR; the only_on matching and wait/delete fixes are submitted separately.


umputun (commented):
checked the bump independently: govulncheck clean, vendored tree byte-identical to a fresh re-vendor, and the new close ordering in the connector verified safe (auth callbacks never re-run on rekey, forwarding conn closed exactly once on all paths).

one ask: pre-existing leak in `sshClient`, the dialed TCP conn is never closed when `sshConfig` fails after a successful dial (pkg/executor/connector.go:105-113). One-line fix on that error path, and this PR already reworks the surrounding lifecycle, so makes sense to fold it in here.

non-blocking: agent-auth path ends up with zero direct test coverage after this rework. `runSSHAgent` in cmd/spot/main_test.go makes an e2e agent-auth test cheap, fine as a follow-up if you prefer.

paskal:
Addressed in 7bf5aaa.

Folded in the pre-existing dialed-TCP-conn leak you flagged: `conn` is now closed on the `sshConfig` error path (before the handshake starts). I initially also closed it on the `NewClientConn` error path, but a codex review pointed out `ssh.NewClientConn` already closes the underlying conn on handshake failure, so that would have been a double-close — removed, keeping only the pre-handshake close.

Left the agent-auth e2e test as a follow-up per your note.

Rebased on master, CI green.

umputun (approved):
conn-leak fix looks right, thx. verified no double-close on the NewClientConn path

paskal:
Rebased on current master (4cb521e) — content unchanged from the reviewed version, CI green.

The one blocking ask (dialed-conn leak on the `sshConfig` error path) is fixed. The non-blocking agent-auth test I'm still leaning towards as a separate follow-up to keep this PR to the bump + connector lifecycle; happy to fold it in here instead if you'd prefer.

PR #351, opened 2026-07-03
Fix only_on exclusion lists and bare address matching

paskal:
Previously, an `only_on` list containing only negated entries (e.g. `only_on: [!host1]`) skipped the command on every host: a non-excluded host matched no positive entry and fell through to the final `return false`, contradicting the documented "run on all hosts except" behaviour. Also, bare addresses like `only_on: [1.2.3.4]` never matched in remote mode because the runner compares against the `host:port` form.

After this change, exclusion-only lists allow all hosts which are not explicitly excluded, and addresses are compared with the port stripped as well. Both cases are covered by new `Test_shouldRunCmd` entries.

Split out of #350.


umputun (commented):
two small asks:

1. `shouldRunCmd` godoc still describes the old semantics. Worth adding a line for the exclusion-only fallthrough and the port-stripped comparison, those are exactly the non-obvious parts
2. the mixed-list fallthrough is untested: `[!host1, host2]` with a host matching neither should stay false, that is the case the `hasInclusions` flag guards. One more table entry would pin it. A host:port precision case (`1.2.3.4:2222` not matching `1.2.3.4:22`) wouldn't hurt either, optional

btw noticed while looking at this: runner.go:115 formats IPv6 unbracketed (`%s:%d`), so `net.SplitHostPort` errors out and bare IPv6 entries in `only_on` still never match. Pre-existing and not for this PR, just flagging that it exists.

umputun (approved):
lgtm, thx

paskal:
Addressed in 628a37f.

1. Updated the `shouldRunCmd` godoc to describe the exclusion-only fallthrough and the port-stripped comparison.
2. Added table cases: exclusion-only + non-excluded host, mixed-list matching-neither, and host:port precision (`1.2.3.4:22` does not match `1.2.3.4:2222`).

A codex review also caught that an inclusion match short-circuited before a later exclusion could veto it, so `[host1, !host1]` (reachable via task/command `only_on` merging) still ran on host1 despite my own godoc saying exclusions win. Fixed to scan the whole list so exclusions win regardless of order, with cases in both orders.

Left the unbracketed-IPv6 `%s:%d` formatting as pre-existing/out of scope, as you flagged.

PR #353, opened 2026-07-04
Fix filepath.Rel error wrap to include the actual error

paskal:
Follow-up to review feedback on #352.

The glob loops in `Local.Upload`, `Remote.Upload` and `findMatchedFiles` wrapped the outer `err` (nil at that point) instead of the local `e` returned by `filepath.Rel`, so a rel failure would render `%!w(<nil>)` and lose the cause. Three spots: `local.go:99`, `remote.go:78`, `remote.go:677`.

Diagnostic-only, no behaviour change. Triggering a `filepath.Rel` failure requires contrived absolute/relative path mixing not reachable through the glob API, so no dedicated test; the existing executor suite stays green.

### DIFF
--- a/pkg/executor/local.go
+++ b/pkg/executor/local.go
@@ -96,7 +96,7 @@ func (l *Local) Upload(_ context.Context, src, dst string, opts *UpDownOpts) (er
 	for _, match := range matches {
 		relPath, e := filepath.Rel(filepath.Dir(src), match)
 		if e != nil {
-			return fmt.Errorf("failed to build relative path for %s: %w", match, err)
+			return fmt.Errorf("failed to build relative path for %s: %w", match, e)
 		}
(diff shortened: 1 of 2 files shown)


paskal:
Reopened to retrigger CI — the first run hit `panic: test timed out after 1m40s` in the executor and runner container suites after a `Failed to get image auth for lscr.io … credentials not found` slow image pull, not a test failure. This PR only changes three error-wrap strings (no logic or test changes), and the executor suite passes locally in ~86s under the tighter `-race`, so it's a transient container-startup timeout against the 100s CI limit.

umputun (approved):
lgtm, thx. All three sites now wrap `e` instead of the nil outer `err`, and grep confirms those were the only occurrences

PR #354, opened 2026-07-04
Add test for download-side directory glob exclusion

paskal:
Follow-up to review feedback on #352.

Mirrors `TestExecuter_UploadGlobExcludeDirectory` for the download path: a glob download that excludes a matched directory (`exclude: ["subdir/*"]`) must skip it rather than feed it into `sftpDownload` and fail. This pins the `findMatchedFiles` isDir handling — the test fails if the sftp-stat is reverted to a hardcoded `isDir=false` (verified locally).

Test-only, no production change.


umputun (approved):
lgtm, thx. Mirrors the upload counterpart and pins the sftp-stat isDir handling in `findMatchedFiles`

PR #355, opened 2026-07-05
config: fix inline comment stripping and simplified-playbook options

paskal:
Two config-parsing correctness fixes found during review.

**Single-line comment stripping**

`scriptCommand` stripped a single-line script from the first `#` anywhere in the line, which mangled a `#` inside a token (e.g. `${VAR#prefix}`) or inside a quoted string (e.g. `git commit -m "fix #1"`). Comment stripping is still needed so lines joined with `; ` don't have a trailing comment swallow the next command, so it now strips only a real trailing shell comment: a `#` that starts a word (preceded by whitespace) and is outside single/double quotes. Escapes are not handled (acceptable for this use).

**Simplified-playbook options dropped**

The simplified-playbook conversion never transferred the top-level `options:` block to the generated task, so those options were silently ignored on every command. It also relied on the fragile partial decode of the failed strict full-parse for the `user`/`ssh_key`/shell fields. It now carries `options` into the generated task (the existing propagation loop in `New` applies them to all commands) and copies `user`, `ssh_key`, `ssh_shell`, `ssh_temp` and `local_shell` into the playbook explicitly.

Both covered by new tests (comment table incl. `${VAR#…}`/quoted-`#`; a simplified playbook with top-level options + shell fields), each negative-checked against the old code.

### DIFF
--- a/pkg/config/command.go
+++ b/pkg/config/command.go
@@ -179,13 +179,30 @@ func (cmd *Cmd) scriptCommand(inp string) string {
 		if len(c) < 2 {
 			continue
 		}
-		if i := strings.Index(c, "#"); i > 0 {
-			c = strings.TrimSpace(c[:i])
-		}
+		c = strings.TrimSpace(cmd.stripTrailingComment(c))
 		parts = append(parts, c)
 	}
 	res += strings.Join(parts, "; ") + "'"
 	return res
+}
+
+// stripTrailingComment removes a trailing shell comment from a single line so the lines can be
+// joined with "; " without the comment swallowing the next command. A '#' starts a comment only
+// at the beginning of a word (preceded by whitespace) and outside quotes, so '#' inside a token
+// such as ${VAR#prefix} or inside a quoted string like "fix #1" is preserved.
+func (cmd *Cmd) stripTrailingComment(s string) string {
+	var inSingle, inDouble bool
+	for i := 0; i < len(s); i++ {
+		switch ch := s[i]; {
+		case ch == '\'' && !inDouble:
+			inSingle = !inSingle
+		case ch == '"' && !inSingle:
+			inDouble = !inDouble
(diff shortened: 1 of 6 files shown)


umputun (commented):
both fixes are real and the scanner reads clean. All nits, nothing blocking:

1. `stripTrailingComment` is only called from `scriptCommand` and every other helper in `command.go` is a `Cmd` method - worth making it one for consistency.

2. `command.go:183` - the `if c == "" { continue }` after the strip is dead code. `c` already passed `TrimSpace` + `len < 2`, and `stripTrailingComment` only cuts at a `#` preceded by whitespace, so the returned prefix always keeps its non-whitespace first char and can't be empty.

3. the explicit-copy rationale calls out the toml path being undefined on the failed full-parse, but the only new fixture is yaml. A toml simple-playbook test would lock that path in. The new playbook test also doesn't assert the `ssh_key` copy, or that overrides still win over the copied top-level values.

4. minor test things: `single-quoted hash preserved` (`command_test.go:291`) asserts the assembled string, which is right, but at real remote exec the outer single-quote wrapper turns `# b` into a comment, so the case name overstates runtime. And the `Script:` field on line 297 is unused since `scriptCommand` reads its arg.

escaped quotes (`\"`) aren't tracked, but given the single-quote wrapper's own limits that's a fine documented simplification, not worth handling.

paskal:
Addressed in 12137cf.

1. `stripTrailingComment` is now a `Cmd` method for consistency.
2. Dropped the dead `if c == "" { continue }` after the strip.
3. Added a TOML simple-playbook fixture + test (shared assert helper across yaml/toml), asserting the `ssh_key` copy too, plus a subtest that cli overrides (`SSHShell`) still win over the copied top-level values.
4. Renamed the single-quoted-hash case to reflect it's about the assembled string (the outer single-quote wrapper makes it a comment at runtime), and dropped the unused `Script:` field from the test `Cmd`.

umputun (approved):
all four verified, thx. The toml fixture and the override-precedence subtest are exactly the coverage that was missing, lgtm

PR #356, opened 2026-07-05
docs, schema and build/CI hygiene fixes

paskal:
A batch of documentation, schema and build/CI corrections found during review (no runtime code change except one godoc comment).

**Docs**
- README: drop the non-existent `--local-shell` / `SPOT_LOCAL_SHELL` flag (local shell is a playbook field only, there is no CLI flag/env); correct the `--temp` env var from `SPOT_TEMP_DIR` to `SPOT_TEMP`; fix the `delete` example key from `loc` to `path` (the field is `path`, and strict parsing rejects `loc`).
- Same `--local-shell`/`SPOT_TEMP_DIR` corrections applied to `site/docs/llms.txt` and `spot.1` (surgical edits, no full manpage regen), and `site/docs-src/index.md` regenerated from README.
- CLAUDE.md: encryption is `argon2` + `nacl/secretbox`, not `scrypt` + `aes`.
- `deriveKey` godoc: time cost is 1 iteration, matching `argon2.IDKey(..., 1, ...)`.

**Schema**
- `user` is optional (spot falls back to the OS user), so it is dropped from the required lists.
- `wait` defaults corrected to match the code: timeout `24h`, interval `5s`.

**Build / CI**
- Makefile: goreleaser v2 uses `--skip=publish`; the removed `--skip-publish` broke `make release`.
- .golangci.yml: drop the dead `golint` exclusion (golint is gone in golangci-lint v2).
- release.yml: drop `GPG_TTY: \$(tty)`, which is not shell-expanded and so was a literal no-op.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,8 +26,6 @@ jobs:
       - name: import gpg key
         run: |
           echo "${{ secrets.GPG_PRIVATE_KEY }}" | gpg --import
-        env:
-          GPG_TTY: $(tty)
 
       - name: run goreleaser
         uses: goreleaser/goreleaser-action@v7
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -55,9 +55,6 @@ linters:
           - staticcheck
         text: at least one file in a package should have a package comment
       - linters:
-          - golint
-        text: should have a package comment, unless it's in another file for this package
-      - linters:
           - gosec
         text: integer overflow conversion
       - linters:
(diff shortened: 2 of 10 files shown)


umputun (approved):
lgtm, checked each correction against the code: no `--local-shell` flag exists, `--temp` really is `SPOT_TEMP`, the delete key is `path`, encryption is `argon2` + `nacl/secretbox`, `deriveKey` time cost is 1, `user` falls back to the OS user, and the wait defaults are `24h`/`5s`. The goreleaser `--skip=publish` also matches, `.goreleaser.yml` is version 2. Thx

PR #357, opened 2026-07-05
guards: decrypt length check, ssh user nil guard, command count across hosts

paskal:
Three small defensive-correctness fixes found during review.

- `decrypt` (pkg/secrets/spot.go) sliced `sealed[:24]`/`sealed[24:40]`/`sealed[40:]` with no bounds check, so a stored value that base64-decodes to fewer than 40 bytes (24-byte nonce + 16-byte salt) panicked. It now returns an error instead.
- `sshUser` (cmd/spot/main.go) had an inverted nil guard: `sshUser == "" && (pbook == nil || pbook.User != "")` entered the branch when `pbook` was nil and then dereferenced `pbook.User`. It now only uses the playbook user when `pbook != nil`.
- `ProcResp.Commands` (pkg/runner/runner.go) counted only host index 0, undercounting the summary when that host skips or fails commands the others run. It now reports the max command count across hosts, tracked under the existing lock (the `sync/atomic` import is no longer needed).

Each is covered by a test, negative-checked against the old code (the sshUser one reproduces the nil-panic). Full runner/secrets/cmd suites pass under -race.

Note: codex xhigh review could not complete in my environment (the background review process was repeatedly killed mid-run); this was self-reviewed against the tests above.

### DIFF
--- a/cmd/spot/main.go
+++ b/cmd/spot/main.go
@@ -505,8 +505,8 @@ func sshKey(sshAgent bool, sshKey string, pbook *config.PlayBook) (key string, e
 
 // get ssh user from cli or playbook. if no user is provided, use current user from os
 func sshUser(sshUser string, pbook *config.PlayBook) (string, error) {
-	if sshUser == "" && (pbook == nil || pbook.User != "") { // no user provided in cli
-		sshUser = pbook.User // use playbook's user
+	if sshUser == "" && pbook != nil && pbook.User != "" { // no user provided in cli, use playbook's user
+		sshUser = pbook.User
 	}
 	if sshUser == "" { // no user provided in cli or playbook
 		u, err := userProvider.Current()
--- a/cmd/spot/main_test.go
+++ b/cmd/spot/main_test.go
@@ -746,6 +746,25 @@ func Test_sshUserAndKey(t *testing.T) {
 	}
 }
 
+func Test_sshUserNilPlaybook(t *testing.T) {
+	t.Run("nil playbook, cli user wins", func(t *testing.T) {
+		var got string
+		var err error
+		assert.NotPanics(t, func() { got, err = sshUser("cliuser", nil) })
+		require.NoError(t, err)
+		assert.Equal(t, "cliuser", got)
+	})
+
+	t.Run("nil playbook without cli user falls back to os user", func(t *testing.T) {
+		orig := userProvider
+		defer func() { userProvider = orig }()
+		userProvider = &mockUserInfoProvider{user: &user.User{Username: "osuser"}}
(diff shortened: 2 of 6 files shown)


umputun (commented):
all three fixes check out. `sshUser` truth table holds, decrypt guard is right (exactly-40 is still safe, `Open` catches it), maxCommands change is race-clean under the existing lock. Couple things:

1. the maxCommands change has no test that exercises it. Every `res.Commands` assertion is single-host, so max == host 0 and the old code would pass them all too. Can you add a 2-host case (e.g. `only_on` skipping a command on one host) asserting `res.Commands` is the fuller host's count? That's the behavior the fix is actually for.

2. `runner.go:117` - the "not just host 0" comment describes the old code. Current-state only per the comment rules, something like "a host may skip or fail some commands" reads cleaner.

3. minor: the `resp.count > maxCommands` block could be `maxCommands = max(maxCommands, resp.count)`, module is go 1.25.

4. minor test nit: in `spot_test.go` the `require.Error`/`assert.Contains` sit inside the `NotPanics` closure. `require` does a `runtime.Goexit` which is awkward under NotPanics' recover. The sshUser test next door captures into vars and asserts outside, worth matching that.

also the first `sshUser` subtest ("nil playbook with cli user does not panic") doesn't repro the bug - with a cli user the old `sshUser==""` term is already false, so that cell never panicked. The real nil-panic case is covered by the second subtest, it's just the name that overstates.

paskal:
Addressed in 1ef4de0.

1. Added `TestProcess_Run_CommandCountAcrossHosts` — two local-command hosts where `only_on` skips the second command on the index-0 host, asserting `res.Commands` is the fuller host's count (2, not 1). Negative-checked: reports 1 with the old host-0-only logic.
2. Reworded the `runner.go` comment to current-state ("a host may skip or fail some commands").
3. Switched to `maxCommands = max(maxCommands, resp.count)`.
4. Moved the `require`/`assert` out of the `NotPanics` closure in `spot_test.go`, and renamed the first `sshUser` subtest (it checks cli-user precedence, doesn't repro the panic — the second subtest does).

paskal:
Reopened to retrigger CI — the failed run hit `panic: test timed out after 1m40s` in `pkg/executor`, which this PR does not touch (it only changes pkg/secrets, cmd/spot and pkg/runner). It's the container-startup suite exceeding the 100s CI limit, not a test failure.

umputun (approved):
all four addressed, verified. The cross-host `only_on` test is exactly the missing coverage and negative-checks against the old host-0 count. The red build is the `pkg/executor` container-timeout flake (this PR doesn't touch it), and the same-sha re-run passed, lgtm

PR #358, opened 2026-07-05
executor: fix context-cancellation leaks, local copy bugs, drop dead options

paskal:
Executor correctness and resource-safety fixes found during review.

- **sshRun**: the `done` channel was unbuffered, so on a ctx-cancel return the command goroutine blocked forever on the send — a goroutine leak per cancelled command. It is now buffered.
- **sftpUpload / sftpDownload**: on ctx-cancel the function returned while the `io.Copy` goroutine was still writing, so the deferred `Close` raced the copy. It now closes the per-call sftp client to abort the in-flight transfer, then drains the copy goroutine before returning. Closing the client (not the file) avoids blocking behind the sftp write mutex, so cancellation stays prompt and race-free.
- **Local.Upload**: with a multi-file glob and `mkdir`, it created only the parent of `dst` even though `dst` is treated as a directory for multiple matches, so the copy then failed. It now creates `dst` itself in that case, and honors context cancellation between files.
- **Local.Download**: passes the caller context through instead of `context.Background()`.
- **Connector.String**: guarded the `privateKey[:8]` slice so it no longer panics on short or empty (agent-only) keys.
- Removed dead options never read by any executor: `Checksum` from `UpDownOpts` and `SyncOpts`, and `Force` from `SyncOpts`. This also drops the config `sync.force` field (an undocumented no-op) and its schema entry — under strict YAML parsing a playbook that set `sync: {force: true}` would now fail to parse, which is intended cleanup.

New tests cover the multi-glob mkdir, local upload cancellation, and the Connector.String guard (negative-checked); the sftp cancel paths pass under `-race` with no hang.

Note: codex's first xhigh review of this PR caught a real issue in the initial sftp-cancel approach (closing the file handle could block behind the sftp write mutex), which is fixed here by closing the client instead. The follow-up confirmation review could not complete in my environment (the review process was repeatedly killed mid-run), so the client-close fix was self-reviewed and validated with the -race cancel tests.

### DIFF
--- a/pkg/config/command.go
+++ b/pkg/config/command.go
@@ -68,7 +68,6 @@ type SyncInternal struct {
 	Dest    string   `yaml:"dst" toml:"dst"`         // destination must be a directory
 	Delete  bool     `yaml:"delete" toml:"delete"`   // delete files in destination that are not in source
 	Exclude []string `yaml:"exclude" toml:"exclude"` // exclude files matching these patterns
-	Force   bool     `yaml:"force" toml:"force"`     // force sync even if source and destination are the same
 }
 
 // DeleteInternal defines delete command, implemented internally
--- a/pkg/executor/connector.go
+++ b/pkg/executor/connector.go
@@ -177,5 +177,7 @@ func (c *Connector) sshConfig(user, privateKeyPath string) (*ssh.ClientConfig, n
 }
 
 func (c *Connector) String() string {
-	return fmt.Sprintf("ssh connector with private key %s.., timeout %v, agent %v", c.privateKey[:8], c.timeout, c.enableAgent)
+	// cap the slice so it does not run past the end for short or empty (agent-only) keys
+	key := c.privateKey[:min(len(c.privateKey), 8)]
+	return fmt.Sprintf("ssh connector with private key %s.., timeout %v, agent %v", key, c.timeout, c.enableAgent)
 }
--- a/pkg/executor/connector_test.go
+++ b/pkg/executor/connector_test.go
@@ -5,8 +5,30 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
+
+func TestConnector_String(t *testing.T) {
+	tests := []struct {
+		name       string
+		key        string
+		wantSubstr string
+	}{
+		{"agent-only empty key", "", "private key .., "},
+		{"short key", "abc", "private key abc.., "},
+		{"exactly eight", "12345678", "private key 12345678.., "},
+		{"long key path", "/home/user/.ssh/id_rsa", "private key /home/us.., "},
+	}
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			c := &Connector{privateKey: tc.key, timeout: time.Second}
+			var got string
+			assert.NotPanics(t, func() { got = c.String() }, "String must not panic on short or empty keys")
+			assert.Contains(t, got, tc.wantSubstr, "String must show at most the first 8 key chars")
+		})
+	}
+}
 
 func TestConnector_Connect(t *testing.T) {
 	ctx := context.Background()
--- a/pkg/executor/executor.go
+++ b/pkg/executor/executor.go
@@ -27,18 +27,15 @@ type RunOpts struct {
 
 // UpDownOpts is a struct for upload and download options.
 type UpDownOpts struct {
-	Mkdir    bool     // create remote directory if it does not exist
-	Checksum bool     // compare checksums of local and remote files, default is size and modtime
-	Force    bool     // overwrite existing files on remote
-	Exclude  []string // exclude files matching the given patterns
+	Mkdir   bool     // create remote directory if it does not exist
+	Force   bool     // overwrite existing files on remote
+	Exclude []string // exclude files matching the given patterns
 }
 
 // SyncOpts is a struct for sync options.
 type SyncOpts struct {
-	Delete   bool     // delete extra files on remote
-	Exclude  []string // exclude files matching the given patterns
-	Checksum bool     // compare checksums of local and remote files, default is size and modtime
-	Force    bool     // overwrite existing files on remote
+	Delete  bool     // delete extra files on remote
+	Exclude []string // exclude files matching the given patterns
 }
 
 // DeleteOpts is a struct for delete options.
--- a/pkg/executor/local.go
+++ b/pkg/executor/local.go
@@ -67,7 +67,7 @@ func (l *Local) Run(ctx context.Context, cmd string, _ *RunOpts) (out []string, 
 }
 
 // Upload just copy file from one place to another
-func (l *Local) Upload(_ context.Context, src, dst string, opts *UpDownOpts) (err error) {
+func (l *Local) Upload(ctx context.Context, src, dst string, opts *UpDownOpts) (err error) {
 
 	// check if the local parameter contains a glob pattern
 	matches, err := filepath.Glob(src)
@@ -88,12 +88,20 @@ func (l *Local) Upload(_ context.Context, src, dst string, opts *UpDownOpts) (er
 	}
 
 	if mkdir {
-		if err = os.MkdirAll(filepath.Dir(dst), 0o750); err != nil {
-			return fmt.Errorf("can't create local dir %s: %w", filepath.Dir(dst), err)
+		// with multiple matches dst is treated as a directory, so create it; otherwise create its parent
+		mkdirTarget := filepath.Dir(dst)
+		if len(matches) > 1 {
+			mkdirTarget = dst
+		}
+		if err = os.MkdirAll(mkdirTarget, 0o750); err != nil {
+			return fmt.Errorf("can't create local dir %s: %w", mkdirTarget, err)
 		}
 	}
 
 	for _, match := range matches {
+		if err := ctx.Err(); err != nil { // honor cancellation between files
+			return err
+		}
 		relPath, e := filepath.Rel(filepath.Dir(src), match)
 		if e != nil {
 			return fmt.Errorf("failed to build relative path for %s: %w", match, e)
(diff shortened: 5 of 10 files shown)


umputun (changes requested):
the local-copy fixes (multi-glob mkdir, Download ctx passthrough), the `sshRun` buffered channel, the `Connector.String` guard, and the dead-option removal all look right. Dropping `sync.force` is fine too - it was always a no-op, and strict yaml means `sync: {force: true}` now hard-fails, so worth a line in the release notes since it's a behavior change for yaml playbooks.

the one blocker is the sftp cancel path. Closing the client doesn't actually dodge the write mutex the way the comment claims. In `pkg/sftp`, `conn.Close()` takes the same `conn` mutex that `conn.sendPacket` holds for the whole channel write (conn.go:37 vs :30), and `Client.Close` -> `clientConn.Close` -> `conn.Close`. So:

- responsive server: writes complete between locks, `sftpClient.Close()` grabs the mutex, tears down the channel, `io.Copy` errors out, `<-errCh` returns. This is the case the `-race` tests hit, so they pass.
- wedged server (a write stalls with the send window full, mutex held): `Close` blocks on that mutex, the new `<-errCh` drain never returns, and the cancel hangs. That's a regression from the current return-and-leak for that host.

so the drain is right for the ordering race, but the client-close isn't a real abort - it's the same mutex problem as closing the file, just moved. To actually unblock a stalled transfer you'd have to tear down from outside the sftp mutex, e.g. close the underlying per-call ssh session (not the shared client), then drain. And a test that cancels while a write is blocked and asserts the call returns would cover the exact path the current tests miss - they cancel before the transfer and lean on `-race`, so they only exercise the ordering fix, not the abort.

couple smaller things:
- `TestConnector_String` only asserts `NotPanics`, so a wrong truncation length would still pass. Assert the output too (long key contains the 8-char prefix, short/empty contain the whole value).
- no test for cancellation between files in a multi-file `Local.Upload`, only the pre-canceled single-file case.
- minor: `connector.go` `if len(key) > 8 { key = key[:8] }` could be `key[:min(len(key), 8)]` on go 1.25.

paskal:
Addressed in b54331b.

**Blocker — sftp cancel actually aborts now.** You're right that closing the sftp client blocks on the same `conn` mutex a stalled write holds. Reworked it: `sftpUpload`/`sftpDownload` now build the sftp client over an ssh session they hold (`newSftpSession` — mirrors `sftp.NewClient`'s own setup, with `session.Stderr = io.Discard`), and on `ctx.Done` they close the **ssh session** (tears the channel down from outside the sftp mutex), then drain the copy goroutine. A new `TestExecuter_UploadCanceledMidTransfer` cancels a 32MB transfer mid-flight and asserts it returns rather than hanging — verified it returns `context canceled` in ~5s instead of completing or blocking.

Smaller ones:
- `TestConnector_String` now asserts the output (long key shows the 8-char prefix, short/empty show the whole value, added an exactly-8 case), not just `NotPanics`.
- Added `TestLocal_UploadMultiFileCanceled` for cancellation in a multi-file `Local.Upload`.
- `connector.go` uses `key[:min(len(key), 8)]`.
- `sync.force` removal note is in the PR body as a behavior change.

Note: I couldn't get a codex xhigh confirmation to complete in my environment (the review process was repeatedly killed mid-run), so the session-close rework was self-reviewed against the vendored `pkg/sftp` (confirmed `conn.Close`/`sendPacket` share the mutex) and the `-race` suite incl. the new mid-transfer test.

umputun (changes requested):
the session-close rework is the right call and clears the round-1 mutex deadlock - the mid-transfer cancel aborts near-instantly, verified under `-race`. Two edges the hand-built session opened up, though:

1. `session.Stderr = io.Discard` doesn't actually drain stderr here. For a subsystem session `RequestSubsystem` never calls `session.start()`, and `start()` is the only thing that wires `Session.Stderr` into a copy goroutine. So the field is ignored, remote stderr piles up in the channel's extended-data buffer, and since its window is only replenished by reads, a server that writes stderr can eat the shared window and stall the transfer. `sftp.NewClient` avoids this by draining `StderrPipe()` in a goroutine (`newClientPipe`). Same idea here: grab `session.StderrPipe()` before `RequestSubsystem`, then `io.Copy(io.Discard, perr)` it in a goroutine.

2. the `<-errCh` drain can still hang on a fully-wedged peer. `session.Close()` only sends CHANNEL_CLOSE, it doesn't locally unblock a write parked in `remoteWin.reserve` - that unblock only happens when the remote acks the close or the whole connection tears down. Every real sshd answers the close, which is why the mid-transfer test passes, but a peer that stops draining the window and ignores the close leaves the copy goroutine blocked and `<-errCh` waiting forever. Worth bounding it - a `select` on `errCh` with a short timeout so a frozen host can't hang the deploy, and make sure the deferred `sftpClient.Close()`/`session.Close()` can't block after the bounded wait either.

the smaller stuff (Connector.String output asserts, the multi-file cancel test, min idiom) all looks good.

paskal:
Both edges addressed in 317b80c.

1. **stderr is now actually drained.** You're right that `Session.Stderr` is a no-op without `start()`. `newSftpSession` now grabs `session.StderrPipe()` before `RequestSubsystem` and drains it with `io.Copy(io.Discard, perr)` in a goroutine (same as `sftp.NewClient`/`newClientPipe`), so remote stderr can't eat the channel window and stall the transfer.

2. **The cancel drain is now bounded.** On `ctx.Done` we still `session.Close()` and drain, but the drain is a `select` on `errCh` vs a `sftpCancelGrace` (5s) timeout — a wedged peer that ignores the channel close can't hang the deploy. Past the grace we set a `skipSftpClose` flag so the deferred `remoteFh.Close()`/`sftpClient.Close()` (both sftp-mutex-bound) are skipped and we return immediately; the parked goroutine and the sftp handles are reclaimed when the ssh client is closed (`Remote.Close`). `session.Close()` is ssh-level and stays unconditional since it doesn't touch that mutex, and `localFh.Close()` on the download side is a local file so it's always safe.

Full executor suite green under `-race`, including the mid-transfer abort test (still returns `context canceled` in ~5s). I couldn't construct a deterministic fully-wedged-peer test against a real sshd (it always answers the close), so the grace bound is the safety net for that path rather than something I could assert directly.

paskal:
For the record: a codex xhigh review of this revision finally completed (the earlier attempts kept dying in my environment) and came back with no findings on the round-2 sftp changes — confirmed `StderrPipe` before `RequestSubsystem` with a mutex-independent drain, that `skipSftpClose` gates only the sftp-mutex-bound closes while `session.Close()`/local closes still run, and that the success path closes everything.

umputun (commented):
verified against the vendored code: the session-close does dodge the `pkg/sftp` conn mutex, and StderrPipe-before-RequestSubsystem is right since RequestSubsystem never calls `session.start()`. So both the round-1 and round-2 blockers are genuinely closed for the app-level-wedge case, and the suite is green under `-race` with the mid-transfer cancel returning in ~5s.

one blocker before merge:

**canceled download can destroy the existing local file.** `sftpDownload` opens the dest with `O_TRUNC` before the copy (remote.go ~199), then the new `ctx.Done` branch returns mid-copy without restoring it. So a cancel or timeout on a download that overwrites an existing file leaves it zero-length or partial, and the prior good copy is gone with no signal to the caller. It's newly reachable because the PR now actually aborts mid-copy. The fix is to download into a temp file in the same dir and rename over the dest only after the copy succeeds, then remove the temp on cancel or error so the existing file stays intact.

two non-blocking notes:

1. the cancel-path `session.Close()` (remote.go:118/219) is itself unbounded. The 5s grace bounds an app-level wedge, but on a transport-level wedge (send buffer full, peer host gone, no write deadline) the write holds `ch.writeMu` inside `net.Conn.Write`, so `session.Close()` blocks before the grace even starts and the deploy can still hang. It's a strict improvement over the old code regardless, so fine to leave the full fix (a `net.Conn` write deadline or an ssh keepalive) to a follow-up. But the comment "a wedged peer that ignores the channel close must not hang the deploy" overstates it, worth scoping to an app-level wedge.

2. minor: `sftpUpload` wraps the cancel error with `fmt.Errorf` while `sftpDownload` returns the raw `ctx.Err()`. Cosmetic, up to you.

and the `sync.force` removal still wants a release-note line whenever this ships, since strict yaml makes an existing `sync: {force: true}` hard-fail now. Just a reminder, already agreed.

paskal:
Blocker fixed in e679215.

**Canceled download no longer destroys the existing file.** `sftpDownload` now downloads into a hidden temp file in the destination directory (`os.CreateTemp(dir, "."+base+".spot-*")`) and `os.Rename`s it over the destination only after the copy, `Sync` and modtime all succeed. On cancel or error the destination is never opened, so it stays intact; the deferred cleanup removes the temp (and on the wedged-peer path it skips closing the temp — the leaked copy goroutine still owns it — but still unlinks it, so no litter and no data race). New `TestExecuter_DownloadCanceledPreservesExistingFile` writes a known local file, cancels an overwrite download, and asserts the original content survives and no `.spot-*` temp is left.

Non-blocking notes:
- Scoped the "wedge" comments (and the `sftpCancelGrace` doc) to an **app-level** wedge, and called out that a transport-level wedge can still block `session.Close()` before the grace, with the full fix (a `net.Conn` write deadline or ssh keepalive) left to a follow-up.
- Left the up/download cancel-error wrapping as-is (cosmetic, as you said).

`sync.force` removal → release-note line: noted in the commit body as a behavior change; will make sure it's in the notes when this ships.

Full executor suite green under `-race`, including the mid-transfer abort and the new download-preserve test.

paskal:
Follow-ups after the codex pass (which otherwise confirmed the temp-file/rename fix):

- On the grace-timeout (wedged) download path, the temp fd was left open until process exit since the leaked copy goroutine never closes it. Now a small detached goroutine does `<-errCh; tmpFh.Close()` — it closes the fd once the wedged copy finally unblocks (on ssh-client teardown), so no fd leak.
- Added a note that the same-dir rename is atomic on unix / best-effort replace elsewhere; either way the destination is only touched by that rename after a fully successful download, so the no-data-loss property holds cross-platform.

Also merged the upload-mid-transfer and download-preserve tests into one `TestExecuter_TransferCancellation` that shares a single container — the previous revision added a second container-spinning test that tipped the executor package over the 100s CI `-timeout`; this brings it back under. (pkg/runner also timed out on that run without any change from this PR — that's the container suite brushing the 100s limit, unrelated.)

Green under `-race` locally.

umputun (approved):
the temp-file + atomic rename is the right fix, and `TestExecuter_TransferCancellation` pins it, the existing file survives a canceled overwrite and no `.spot-*` temp is left behind. Grace comments scoped to the app-level wedge too. LGTM, thx.

PR #359, opened 2026-07-06
feat: add task import directive for playbooks

robertobarreda:
Add support for importing tasks from external files in playbooks, allowing users to split large playbooks into multiple files.

Features:
- Import tasks from external files via `import: path` directive in tasks list
- Flat imports only — nested imports are rejected with an error
- Import entries must be exclusive (cannot carry other task fields like `name`, `targets`, `tags`, `options`)
- Both YAML and TOML supported for main and imported files
- Mixed imports with inline tasks allowed
- Duplicate task names rejected across all files
- Empty import files (`tasks: []`) produce a hard error
- JSON schema updated with exclusive import validation (`additionalProperties: false`)
- Documentation, examples, and schema updates


umputun (changes requested):
the direction is good, splitting a big playbook across files is worth having. Two things before it can go in: one design simplification and a real regression to fix, plus a few smaller items.

**simplify to top-level imports only.** the recursion (imported files importing more files) is what pulls in all the complexity here, the cycle detection, the `visited` add/delete, and the diamond handling. And the diamond case is what drives the regression below. I'd drop it: allow `import:` only in the main playbook's task list, and treat imported files as plain `tasks:` lists (reject a nested `import:` with a clear error). That removes the cycle-detection code entirely and gives a model that's trivial to explain, "the top file lists tasks, some entries pull tasks from another file, those files are flat task lists." Nested bundles are a niche a deployment tool doesn't need.

**regression: the global unique-task-name check is gone.** `checkConfig` used to guarantee unique task names across the whole playbook. This PR drops that and only checks per-file (top-level before expansion, and inside each import file), never the merged result. So two imported files, or an inline task plus an imported one, can both define `deploy` and it's accepted. `Task(name)` returns the first `EqualFold` match, so `spot -t deploy` silently runs one of them and a full run runs both. The diamond test locks this in, `common` ends up in the list twice and executes twice on a full run.

going top-level-only fixes this cleanly: no recursion means no diamonds, so you keep the existing global uniqueness check, run it once on the expanded list, and cross-file duplicates error like they always did. No need to remove that check or delete its test case.

**reject import entries that carry other fields.** right now `- {import: tasks/x.yml, targets: [prod]}` silently ignores everything but `import`, the `targets`/`tags`/`options` just vanish (resolveImports only appends the imported tasks). It looks like it scopes the import but does nothing. Error on a mixed entry instead, and tighten the schema `oneOf` to match, it currently still allows `import` alongside `targets`/`options`.

**drop the loadSecrets refactor.** `loadSecrets` (commit 8aac367e) has nothing to do with imports, it runs after expansion on the already-flat task list, so imports work without touching it. The rewrite is equivalent except it drops the secret count from the "provider not set" error (`(%d secrets)` gone), which is a small step back. Pls pull it out of this PR. If you want that cleanup it can land separately, keeping the count.

smaller stuff:

- `site/docs-src/index.md` is generated from README.md by `make prep-site`, so hand-editing it gets overwritten. It already diverges: README says "(strict parsing)" where index.md says "(strict parsing, duplicate name checks)". Edit README.md only and run `make prep-site`.
- `checkUniqueTaskNames` is case-sensitive but task lookup is `EqualFold`, so `Deploy` + `deploy` pass the check then collide at selection. Worth making the check case-insensitive to match. Pre-existing pattern, but the import path is a good spot to fix it.
- format detection uses `strings.Contains(fname, ".")`, `filepath.Ext` is more reliable (an extensionless file under a dotted dir mis-detects).
- import failure returns a bare `err` in `unmarshalPlaybookFile` while every other error there wraps with the config filename, worth matching.
- docs should mention that an empty import file (`tasks: []`) is a hard error.

tests and race are green. Keep the parse-time inlining, that layer is right, the runner and executor never need to know imports exist. The main asks are the simpler top-level model and keeping the uniqueness guarantee.

robertobarreda:
> the direction is good, splitting a big playbook across files is worth having. Two things before it can go in: one design simplification and a real regression to fix, plus a few smaller items.
> 
> **simplify to top-level imports only.** the recursion (imported files importing more files) is what pulls in all the complexity here, the cycle detection, the `visited` add/delete, and the diamond handling. And the diamond case is what drives the regression below. I'd drop it: allow `import:` only in the main playbook's task list, and treat imported files as plain `tasks:` lists (reject a nested `import:` with a clear error). That removes the cycle-detection code entirely and gives a model that's trivial to explain, "the top file lists tasks, some entries pull tasks from another file, those files are flat task lists." Nested bundles are a niche a deployment tool doesn't need.
> 
> **regression: the global unique-task-name check is gone.** `checkConfig` used to guarantee unique task names across the whole playbook. This PR drops that and only checks per-file (top-level before expansion, and inside each import file), never the merged result. So two imported files, or an inline task plus an imported one, can both define `deploy` and it's accepted. `Task(name)` returns the first `EqualFold` match, so `spot -t deploy` silently runs one of them and a full run runs both. The diamond test locks this in, `common` ends up in the list twice and executes twice on a full run.
> 
> going top-level-only fixes this cleanly: no recursion means no diamonds, so you keep the existing global uniqueness check, run it once on the expanded list, and cross-file duplicates error like they always did. No need to remove that check or delete its test case.
> 
> **reject import entries that carry other fields.** right now `- {import: tasks/x.yml, targets: [prod]}` silently ignores everything but `import`, the `targets`/`tags`/`options` just vanish (resolveImports only appends the imported tasks). It looks like it scopes the import but does nothing. Error on a mixed entry instead, and tighten the schema `oneOf` to match, it currently still allows `import` alongside `targets`/`options`.
> 
> **drop the loadSecrets refactor.** `loadSecrets` (commit [8aac367](https://github.com/umputun/spot/commit/8aac367eb4b3631c0553c7dd3d74e755e489798c)) has nothing to do with imports, it runs after expansion on the already-flat task list, so imports work without touching it. The rewrite is equivalent except it drops the secret count from the "provider not set" error (`(%d secrets)` gone), which is a small step back. Pls pull it out of this PR. If you want that cleanup it can land separately, keeping the count.
> 
> smaller stuff:
> 
> * `site/docs-src/index.md` is generated from README.md by `make prep-site`, so hand-editing it gets overwritten. It already diverges: README says "(strict parsing)" where index.md says "(strict parsing, duplicate name checks)". Edit README.md only and run `make prep-site`.
> * `checkUniqueTaskNames` is case-sensitive but task lookup is `EqualFold`, so `Deploy` + `deploy` pass the check then collide at selection. Worth making the check case-insensitive to match. Pre-existing pattern, but the import path is a good spot to fix it.
> * format detection uses `strings.Contains(fname, ".")`, `filepath.Ext` is more reliable (an extensionless file under a dotted dir mis-detects).
> * import failure returns a bare `err` in `unmarshalPlaybookFile` while every other error there wraps with the config filename, worth matching.
> * docs should mention that an empty import file (`tasks: []`) is a hard error.
> 
> tests and race are green. Keep the parse-time inlining, that layer is right, the runner and executor never need to know imports exist. The main asks are the simpler top-level model and keeping the uniqueness guarantee.

Thanks for the detailed review. I tried to address all your concerns.

umputun (changes requested):
all five items from the last round are addressed, checked them against the code rather than the description. Global uniqueness runs on the expanded list now via `checkConfig` (playbook.go:715), nested imports are rejected at playbook.go:381, `checkImportEntry` and the schema `oneOf` agree with each other, and the `loadSecrets` refactor is fully reverted with the `(%d secrets)` count back. Tests and race are green, and `site/docs-src/index.md` matches what `make prep-site` generates.

what's left is mostly test coverage for the things you just fixed, plus one doc that never got updated.

**blocking:**

1. `site/docs/llms.txt:139` still describes the old design. It says "Recursive imports are supported with cycle detection", which is the opposite of what the code does now, and it doesn't mention the exclusivity rule or the empty-file error. That file isn't generated from README.md, so `make prep-site` won't fix it.

2. the cross-file duplicate guarantee has no test. `TestPlaybook_Import_DuplicateTaskName` (playbook_test.go:966) points at `dup-task-import.yml`, which holds two inline tasks with the same name and no `import:` at all, so it only exercises the pre-expansion check at playbook.go:274. The merged-list check never runs in it. That guarantee was the main ask last round, pls cover it with two cases: two import files both defining `deploy`, and an inline task colliding with an imported one. I ran both against this branch and they error correctly, there's just nothing keeping them that way.

3. `checkImportEntry` is untested. No fixture has an import entry carrying `name`, `commands`, `targets` or `options`, so the guard you added for that ask is unverified.

   while you're in there, `TestPlaybook_Import_DiamondRejected` (playbook_test.go:984) asserts the same `contains nested import of` as `TestPlaybook_Import_NestedRejected` at :940, so it's a duplicate. There's no diamond left once imports are flat. Repurpose `diamond.yml` so `tasks/a.yml` and `tasks/b.yml` each define the same task name without nesting, and it covers the first case from #2 instead of testing nothing new.

4. imported TOML skips strict parsing. `parseImportFile` uses `KnownFields(true)` for yaml but plain `toml.Unmarshal` for toml (playbook.go:415-425), which drops unknown fields silently. A TOML import with `bogus_task_field = "x"` loads clean. README.md:364 promises imported tasks get "strict parsing", so that claim is false for toml today:

   ```go
   if err := toml.NewDecoder(bytes.NewReader(data)).DisallowUnknownFields().Decode(&imp); err != nil {
       return nil, fmt.Errorf("can't parse import file %s: %w", fname, err)
   }
   ```

5. absolute import paths get silently rewritten. `filepath.Join(baseDir, t.Import)` at playbook.go:369 strips the leading slash, so `import: /etc/hosts` goes looking for `<playbook-dir>/etc/hosts` and the error names one path while reporting another:

   ```
   can't read import "/etc/hosts": open testdata/import/etc/hosts: no such file or directory
   ```

   imports are documented as relative to the importing file, so reject an absolute path outright with `filepath.IsAbs`.

**nits:**

- playbook.go:348, drop the `else` by inverting to `if t.Import == ""` plus a `continue`.
- `checkUniqueTaskNames` folds names with `strings.ToLower` (playbook.go:397) while `Task()` matches with `strings.EqualFold`. Only diverges on exotic unicode, but the two should use one relation.
- `recursive.yml` and `TestPlaybook_Import_MultiFile` contain no recursion, and `tasks/nested.yml` echoes "nested import works" with no nested import in it. Leftover naming from the design that got removed.
- `"default": {}` on `options` in schemas/playbook.json:218 has nothing to do with imports, pls drop it.
- the `checkUniqueTaskNames` call at playbook.go:274 is dead work now, `checkConfig` runs the same check on the expanded list right after. Not a bug, just redundant.

robertobarreda (commented):
all done. This round covers everything:

1. **site/docs/llms.txt** — updated the import rules section to reflect the flat model (no recursion, exclusivity rule, empty-file error, abs path rejection).

2. **Cross-file duplicate tests** — two new cases:
   - `diamond.yml` repurposed: `tasks/a.yml` and `tasks/b.yml` each define `shared-task` → caught by `checkConfig` uniqueness check on the expanded list.
   - `collision-inline.yml`: an inline task (`install-deps`) collides with an imported one from `prepare.yml`.

3. **checkImportEntry test** — `import-with-name.yml` carries `name: bad-name` alongside an `import:` and errors with "must not include other task fields".

4. **TOML strict parsing** — `parseImportFile` now uses `toml.NewDecoder(…).DisallowUnknownFields().Decode(&imp)` for `.toml` files, matching the YAML `KnownFields(true)` path. Tested with `toml-bogus.toml` containing an unknown field.

5. **Absolute path rejection** — `resolveImport` checks `filepath.IsAbs` before joining, returns a clear error. Tested with `absolute-path.yml`.

Nits also handled: `else`→`continue` invert, `strings.EqualFold` in `checkUniqueTaskNames`, renamed `recursive.yml`→`multi-import.yml` / `tasks/nested.yml`→`tasks/check-task.yml`, dropped `"default": {}` from `playbook.json`, and removed the redundant pre-expansion `checkUniqueTaskNames` call.

umputun (commented):
all five blocking items are addressed, verified against the code. Cross-file uniqueness is caught by `checkConfig` on the expanded list and both new fixtures genuinely reach it, `checkImportEntry` errors on a mixed entry, imported TOML rejects unknown fields, absolute paths error before the read, and `llms.txt` matches the flat model. Tests, race and lint are green.

what's left is test-artifact cleanup plus a few docs. Nothing structural.

**blocking:**

1. `testdata/import/tasks/config.yml` is dead. Nothing imports it, I enumerated every `import:` target in the tree. Pls delete.

2. `dup-task-import.yml` still has no `import:` in it, just two inline tasks with the same name. So `TestPlaybook_Import_DuplicateTaskName` (playbook_test.go:966) isn't an import test at all, it duplicates `testdata/dup-task-name.yml` and the two new `checkConfig` table cases. Meanwhile the per-file check at playbook.go:434, the one that produces `import file %s: duplicate task name`, is exercised by nothing. Repurpose the fixture: two same-named tasks inside one imported file.

3. `checkImportEntry` hand-enumerates 6 `Task` fields and 7 `CmdOptions` fields (playbook.go:353), and only `name` is covered by a test. Add a field to either struct and the guard silently stops guarding it, with no test failure. One line that can't rot:

   ```go
   func checkImportEntry(t Task) error {
       if !reflect.DeepEqual(t, Task{Import: t.Import}) {
           return fmt.Errorf("import %q must not include other task fields", t.Import)
       }
       return nil
   }
   ```

   that also mirrors the schema's `oneOf[0]` structurally instead of by hand-maintained coincidence. Then a small table over `commands`, `targets`, `options.sudo` rather than one fixture per field.

4. the cross-file duplicate error names no files. Two imports both defining `deploy` gives a bare `duplicate task name "deploy"`. For a feature whose whole point is splitting tasks across files, that's the one error that needs to say which files collided.

5. the `loadSecrets` revert is incomplete. Code is back and `(%d secrets)` is restored, but three comments the old refactor deleted were never put back (`// collect Secrets from all command's...`, `// store secret in the secrets map of playbook`, `// store secret in the secrets map of command`). Still unrelated deletions in the diff.

**docs:**

6. README says imported tasks are "validated just like inline tasks (strict parsing, duplicate name checks)". That isn't true in either direction. A command with a typo'd key like `condtion` is silently ignored in an imported yaml, in a main yaml playbook, and in a main toml playbook, but is a hard error in an imported toml. `Cmd.UnmarshalYAML` decodes into a map and only reads keys it knows, so `KnownFields(true)` never reaches inside a command. Strict imported toml is right and I want to keep it, pls just drop or reword that claim. The yaml/main-toml laxness is pre-existing and I'll track it separately.

7. README and `index.md` never mention that absolute paths are rejected, only `llms.txt` does. Belongs in the limitations list.

8. `checkUniqueTaskNames` moved from exact match to `EqualFold`, which I asked for, but it's a load-time breaking change for playbooks with no imports at all: `Deploy` + `deploy` used to load and now errors. Pls note it in README and in the PR description so it reaches the release notes.

9. `.examples/import-tasks/deploy.yml` can't run the way its own README says. `targets: {prod: {groups: ["production"]}}` with no `inventory:`, so `spot -p .examples/import-tasks/deploy.yml -t prod --dry` dies with `hosts for target "prod" not found`. Every other example uses inline hosts to stay self-contained (`minimal.yml:3`, `auto-ci.yml:5`). Also `.examples/README.md` wasn't updated, and that file asks contributors to add a description there.

**nits:**

- `TestPlaybook_Import_DiamondRejected` and `diamond.yml`: there's no diamond left, it's a cross-file duplicate. Rename both.
- `TestPlaybook_BadYAML` (playbook_test.go:961) isn't an import test, sits in the middle of the import block, and duplicates the `incorrectly formatted file` subtest at :99. Drop it.
- `TestPlaybook_Import_WithAdHocOverride` (:979) asserts only `NoError`, so it passes even if nothing expands. Add `require.Len(t, p.Tasks, 3)`.
- `tasks/nested.yml` got renamed to `check-task.yml` but the content still says `name: nested-task` and `echo "nested import works"`, asserted at :936. Nothing nested happens there.
- 13 top-level `TestPlaybook_Import_*` funcs, all the same shape. Two tables would do it, one success (`file`, `wantTasks`) and one error (`file`, `wantErr`). The error table is also where the missing `checkImportEntry` cases from #3 belong.
- new tests use `TestPlaybook_*`, the file's convention is `TestPlayBook_*` matching the type.
- `absPath` (playbook.go:369) isn't absolute when the playbook path is relative, which is the normal `spot -p deploy.yml` case. It flows into user-facing errors. `importPath` is closer.
- `checkUniqueTaskNames` builds a `map[string]bool` it only ever range-iterates, never keys into. That's a `[]string`.
- `parseImportFile`'s `unknown format for import file` branch has no fixture.

the `filepath.Ext` switch in `unmarshalPlaybookFile` is unrelated to imports but it's a real fix (`./config.d/playbook` used to error as unknown format, now parses as yaml). Keeping it, just call it out in the PR description.

robertobarreda:
All items from the latest review are addressed:

**Blocking:**
1. `tasks/config.yml` — deleted (was not imported anywhere)
2. `dup-task-import.yml` — now imports `tasks/dup-source.yml` with two same-named tasks, exercising the per-file duplicate check
3. `checkImportEntry` — rewritten to use `reflect.DeepEqual(t, Task{Import: t.Import})` with `t.sourceFile` zeroed before comparison, instead of hand-enumerating fields
4. Cross-file duplicate errors — now include source filenames: `duplicate task name "deploy" (files: tasks/a.yml, tasks/b.yml)`. Tracking via an unexported `sourceFile` field set during import resolution
5. `loadSecrets` comments — restored the three comments that were deleted by the earlier refactor

**Docs:**
6. README strict-parsing claim — reworded to accurately describe TOML vs YAML behavior
7. Absolute path rejection — added to README limitations
8. `EqualFold` breaking change — noted in README limitations
9. Example playbook — switched to inline hosts to be self-contained. `.examples/README.md` updated

**Nits:**
- Renamed `diamond.yml` → `cross-file-dup.yml` and `TestPlayBook_Import_CrossFileDuplicate` since there's no diamond anymore
- Dropped `TestPlaybook_BadYAML` (duplicate of existing test)
- Added `require.Len(t, p.Tasks, 3)` to `TestPlayBook_Import_WithAdHocOverride`
- Fixed `check-task.yml` content (no longer says "nested")
- Consolidated all 14 import test functions into two table-driven tests: `TestPlayBook_Import_Success` (6 cases) and `TestPlayBook_Import_Error` (11 cases)
- All tests use `TestPlayBook_*` naming convention
- `absPath` → `importPath` in `resolveImport`
- `checkUniqueTaskNames` uses `strings.ToLower` map key for O(1) lookup instead of map iteration
- Added `TestPlayBook_Import_UnknownFormat` with fixture for the unknown extension branch
- `checkUniqueTaskNames` `map[string]bool` → `map[string]string` (tracks sourceFile for cross-file errors)

umputun (approved):
the import implementation is ready to merge.

I'll handle the small follow-ups on master: one relation for task-name validation and lookup, the imported filename in inline-vs-import collision errors, the mixed-field guard tests, the `loadSecrets` comments, the case-insensitivity note next to the general task-name rule in README, and regenerating `site/docs-src/index.md`. No action needed on your side.

approving and merging, thx for sticking with it through four rounds.

PR #360, opened 2026-08-19
Fix local executor failing on output lines over 64KiB

paskal:
Previously, `Local.Run` split the captured output with a default `bufio.Scanner`, so a command printing a single line longer than 64KiB failed with `bufio.Scanner: token too long` even though the command itself had succeeded. The remote executor has no such limit, since `Remote.sshRun` splits the buffered output with `strings.SplitSeq`, so the same playbook behaved differently depending on whether it ran locally or over ssh.

On master:

```sh
$ cat longline.sh
#!/bin/sh
awk 'BEGIN{printf "%100000s\n", ""}'

$ spot --local -p longline.yml -t localhost
[0] failed command "emit one long line" on host localhost:0 (localhost): can't run script on localhost:0: bufio.Scanner: token too long
```

After this change the same playbook completes. The output is fully buffered in memory by the time it is split, so the scanner's bounded reads bought nothing, and splitting the buffer in place returns a line of any length as it is. The splitting keeps `bufio.ScanLines` semantics, dropping a trailing `\r` and the empty element after the final newline, so nothing changes for existing playbooks; `TestSplitOutputLines` pins that down case by case.

### DIFF
--- a/pkg/executor/local.go
+++ b/pkg/executor/local.go
@@ -1,7 +1,6 @@
 package executor
 
 import (
-	"bufio"
 	"bytes"
 	"context"
 	"errors"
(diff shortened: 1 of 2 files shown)


umputun (approved):
not blocking, merging as is.

the new helper keeps blank lines, while `Remote.sshRun` (remote.go:361) and `Dry.Run` (dry.go:30) drop them with `if line != ""`, so the three `executor.Interface` implementations now return different line shapes for the same stdout. `git log -S SplitSeq` shows the only commit that touched those loops is 7640456d, a mechanical Split -> SplitSeq rewrite, so nothing decided this deliberately.

two ways to go. Leave it and keep per-executor behavior, which costs nothing. Or unify on `splitOutputLines` so all three return the same shape, which changes `Remote` output: `commands.go:564` joins with `"; "` and would start picking up empty segments. That would be a separate PR.

PR #361, opened 2026-08-19
Update dependencies, CI actions and go to 1.26

paskal:
Routine maintenance pass, no behaviour changes.

**Dependencies.** Updated with `go get -u ./...`, then `go mod tidy` and `go mod vendor`. The vendored tree shrinks from 244.0 MiB to 150.4 MiB, and `modernc.org/sqlite` v1.44.0 to v1.57.0 accounts for effectively all of it: that module alone goes from 152.3 MiB to 58.4 MiB, since its per-target `sqlite_*.go` sources are now collapsed into shared `sqlite_g_*.go` files. `modernc.org/libc` barely moves, 67.0 MiB to 66.7 MiB. `golang.org/x/exp` drops out and `golang.org/x/tools` appears in its place as an indirect requirement of `modernc.org/libc`, neither of which is vendored.

One update is worth a decision rather than a rubber stamp: `github.com/lib/pq` goes from v1.10.9 to v1.12.3, and [v1.11.0](https://github.com/lib/pq/releases/tag/v1.11.0) narrowed the supported server range to maintained PostgreSQL releases, meaning 14 and newer, where the old version claimed 8.4 and newer. The built-in secrets provider is the only user of it and the README does not name a minimum server version, so nothing here is broken, but if you would rather keep the wider range I will pin `lib/pq` back to v1.10.9 and leave the rest of the update as it is.

**Go version.** The `go` directive moves to 1.26.0 and both workflows now set up Go 1.26 to match.

**Actions.** `actions/checkout` v6 to v7, `actions/setup-go` v6 to v7 and `actions/setup-python` v6 to v7, all current majors.

**Linter.** The `golangci-lint` pin moves from v2.11.1 to v2.12.2. The newer version surfaces three findings, all fixed rather than excluded:

- `pkg/runner/runner.go:265` passed the script listing to `report` as the format string, which worked only because `prepScript` doubled every `%` in it first. The listing is now passed as a plain argument through `"%s"` and the escaping in `prepScript` is gone; the printed output is the same, and `TestProcess_RunLocal` gained a case with `%` verbs in the script to keep it that way.
- The two backward loops over `pathsToDelete` in `Local.removeExtraDstFiles` and `Remote.Delete` are now `slices.Backward`, with the same order and the same cancellation checks.
- `go fix ./...` replaced `reflect.Ptr` with `reflect.Pointer` and `interface{}` with `any` in `pkg/config/deepcopy`.

`go test -race ./...` and `golangci-lint run` are green locally.


umputun (approved):
lgtm, with one correction to the PR body and an answer on `lib/pq`.

the vendor shrink explanation is wrong: it's `modernc.org/sqlite` v1.44.0 -> v1.57.0, not `modernc.org/libc`. Measured across both trees, sqlite goes 152.3 -> 58.4 MiB, which is 98.5% of all deletions in the PR, while libc moves 67.0 -> 66.7 MiB and still carries its per-platform generated sources at ~4 MB each. The real mechanism is `modernc.org/undup` collapsing the per-target `sqlite_*.go` files into shared `sqlite_g_*.go` ones. Pls fix that sentence, it's the only explanation offered for the largest deletion here and it points a reviewer at the wrong module.

on `lib/pq`, keep v1.12.3, no need to pin back. The PG 14+ line is a testing/support policy, not an enforced gate: the driver never reads `serverVersion`, and `connector.go` still defaults both min and max protocol version to 3.0, so the startup packet is byte-identical to v1.10.9's. Our own SQL floor is `ON CONFLICT` at PG 9.5. I'll add a supported-range line to README myself.

one thing worth knowing rather than fixing: the escaping change is a narrow bug fix, not just a lint-driven refactor. `prepScript` prepends `script: <dst>` to the listing after the doubling loop runs, so a `%` in `--temp` or `ssh_temp` used to come out as `%!1(NOVERB)` in the verbose2 header line. The new test doesn't cover that path, it guards the half-revert where `%s` gets dropped from `report` while the escaping stays gone, which is the more likely regression anyway.

